PR #27, opened 2021-05-14
UK Carbon Instensity Statistics (stats) to a excel file

AdamSulek:
StatsToExcel class may propably write inside StatsToCSV but i create another file. 
Also not sure if to_excel function should be in  Source instead of UKCarbonIntensity class but in my case it works. 
Jupyter dose not open excel file so after revoke run(), please open xlsx file from parent directory.


acivitillo:
@AdamSulek the code looks good, can you also add a unit test here: https://github.com/dyvenia/viadot/blob/main/tests/prefect/test_uk_carbon_intensity_tasks.py

You should also run those tests, try `pytest test_uk_carbon_intensity_tasks.py`. If this doesn't work it's the right time to start using docker as viadot's docker has the correct libraries. Check "Running Tests" in the README file of viadot.

The code is a bit weird, you need to understand what a `fixture` is (line 9). You can ping me today, @Trymzet is an expert of fixtures and he is working with me today too.

AdamSulek:
Tests for StatsToCSV and StatsToExcel class. 
Documentation for using StatsToCSV and StatsToExcel class. 
Library for StatsToCSV and StatsToExcel class.

.config/credentials.json.template
```diff
@@ -1,23 +0,0 @@
-{
```
acivitillo, line 1: @AdamSulek your commit is eliminating this file from the `main` branch. Please uncommit this file from your PR (or just add it back and commit again). In general be careful using `git commit add .` as it adds all your repo changes, it is better to add folders or files `git commit add tests`

tests/prefect/test_uk_carbon_intensity_tasks.py
```diff
@@ -1,17 +1,40 @@
...
+
+def test_uk_carbon_intensity_to_excel(ukci_task_excel):
+    ukci_task_excel.run(path=TEST_FILE_PATH_EXCEL)
+    if_exist = os.path.isfile(TEST_FILE_PATH_EXCEL)
+    assert if_exist == True
+
+def test_uk_carbon_intensity_to_excel_contain(ukci_task_excel):
+    ukci_task_excel.run(path=TEST_FILE_PATH_EXCEL)
+    excel_file = load_workbook(TEST_FILE_PATH_EXCEL)
+    excel_file2 = excel_file.active
```
acivitillo, line 38: you don't need `excel_file2`, we normally don't use names like `var1`. You should be able to do `excel_file = load_workbook(TEST_FILE_PATH_EXCEL).active`

acivitillo (changes requested):
@AdamSulek thanks for the PR, it looks good, I have only requested 2 additional changes. If you update your PR I can approve it and you have your first commit in /main!

AdamSulek:
Adding a config.json.template 
Changing test for excel file with uk_carbon_intensity with excel file content

PR #34, opened 2021-05-21
Creating SQlite database from pandas DataFrame and getting pd.DataFrame from sql query

AdamSulek:
Main changes in base (Source) - insert method previously now really insert values into Table so I change it.
Also I write new method in base -  to_df method but I am not sure where it should be? Maybe in SQLite class?

LoadDF(Task) - Task which create SQL database from pandas DataFrame

RunSQL(Task) - Task which generate pandas DataFrame from sql query

DOCS and tests


tests/prefect/test_tasks_sqlite.py
```diff
@@ -0,0 +1,35 @@
...
+    load_table = LoadDF()
+    yield load_table
+
+
+def test_create_table_from_df(load_table):
+    dtypes = {"country": "VARCHAR(100)", "sales": "FLOAT(24)"}
+    data = defaultdict(list)
+    data["country"].append('italy')
+    data["sales"].append(100.0)
+    df_data = pd.DataFrame(data)
```
acivitillo, line 23: please create `data` as a normal dict of arrays, you don't need to use append, do something like `data = {"sales":[100,200]}`

tests/prefect/test_tasks_sqlite.py
```diff
@@ -0,0 +1,35 @@
...
+    df = load_table.run(table_name=TABLE,
+                   schema=None,
+                   dtypes=dtypes,
+                   db_path=DB_PATH,
+                   df=df_data,
+                   if_exists="replace"
+                   )
+    assert df == True
```
acivitillo, line 31: `df == True` is hard to understand as `df` usually means pandas DataFrame and in this case df is not a DataFrame. Please rename this variable with a more descriptive name

tests/prefect/test_tasks_sqlite.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+import pandas
+from viadot.tasks.sqlite_tasks import RunSQL, LoadDF
+from collections import defaultdict
+import pandas as pd
+import os
+
+TABLE = "test"
+DB_PATH = ':memory:'
```
acivitillo, line 9: what is this for? It seems you are not using this variable in the `LoadDF()` Task class...

viadot/tasks/sqlite_tasks.py
```diff
@@ -0,0 +1,51 @@
...
+            df: pd.DataFrame = None,
+            if_exists: str = "fail"):
+
+        sqlite = SQLite(driver="{SQLite}", server="localhost", db=db_path)
+        sqlite.create_table(table=table_name, schema=schema, dtypes=dtypes, if_exists=if_exists)
+        sqlite.insert_into(table=table_name, df=df)
+
+        return True
+
+class RunSQL(Task):
```
acivitillo, line 32: can you check this task with @Trymzet

I think the code makes sense but since now we are returning a DataFrame we might need to rename it. Something like `SQLtoDF()`. @Trymzet please discuss with @AdamSulek 

Also you need to add a validation rule that the SQL starts with `select` because if it starts with `create` or anything else your sql execution would return no results.

acivitillo (changes requested):
@AdamSulek please review

viadot/tasks/sqlite_tasks.py
Trymzet: Yes, the name doesn't make sense as it seems you're only fetching SELECT results, not running an arbitrary query. If you're running other types of operations you also need to ensure you're committing the transaction. You can check out the query task I added for Azure SQL database on 0.12 branch.

AdamSulek:
Insert (Task) - Task which create SQL database from insert pandas DataFrame

SQLtoDF(Task) - Task which generate pandas DataFrame from sql query

DOCS and tests

mkdocs.yml
```diff
@@ -5,6 +5,7 @@ nav:
       - "Config File": howtos/config_file.md
       - Flows: howtos/flows.md
       - "Get Excel file": howtos/uk_carbon_intensity.md
+      - "Get data as pandas DataFrame": howtos/run_sql.md
```
acivitillo, line 8: Please use a better shorter description than "Get data as pandas DataFrame"

tests/prefect/test_tasks_sqlite.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.fixture(scope="session")
+def load_table():
+    load_table = Insert()
+    yield load_table
+
+
+def test_create_table_from_df(load_table):
+    dtypes = {"country": "VARCHAR(100)", "sales": "FLOAT(24)"}
+    df_data = pd.DataFrame({"country": ['italy'], "sales": [100.]})
+    load = load_table.run(table_name=TABLE,
```
acivitillo, line 19: variable names should be nouns not verbs, use `result`

docs/howtos/run_sql.md
```diff
@@ -0,0 +1,10 @@
+## Loading Data to a Source
```
acivitillo, line 1: can you explain what is db_path and sql_path? You have only 2 sentences here, try to expand a bit the documentation for the `SQLtoDF` task

viadot/sources/base.py
```diff
@@ -122,6 +123,13 @@ def run(self, query: str):
             return cursor.fetchall()
         self.con.commit()
 
+    def to_df(self, query: str):
+        conn = self.con
+        if query.upper().startswith("SELECT"):
+            return pandas.read_sql_query(query, conn)
+        else:
+            return pandas.DataFrame()
```
acivitillo, line 131: you need to write a test for this case

acivitillo (changes requested):

PR #99, opened 2021-08-16
📝 Add local setup section and update contribution guide in README.md 

Lauralicja:
- created a section for local setup (before the run sections
-  How to contribute section update

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,9 +97,10 @@ However, when developing, the easiest way is to use the provided Jupyter Lab con
 
 ## How to contribute
 
-1. Setup locally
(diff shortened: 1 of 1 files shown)


Trymzet (changes requested):
Good catch with the priviledges. Since we're writing the document in the form of a recipe, can you add the exact instructions to do that? Sth like: if you're on Unix, run `sudo chmod +x update.sh` and `sudo chmod +x run.sh`.

ZuchowskaDyvenia:
Done.
Check PR #164

PR #100, opened 2021-08-16
🐛 Fix early return in json_data loop

Lauralicja:
When testing the uk carbon intensity example I found out that it always returns one row.
The bug was caused by an badly indented return which exited the loop too early.

TODO: add info to changelog


Trymzet:
Thanks :) Are you also willing to also add a changelog entry and a test for this case?

Trymzet (changes requested):
TODO: 
- add a changelog entry
- add a test for this case

Trymzet:
LGTM

Issue #209, opened 2021-11-26
CloudForCustomersReportToADLS append doesn't work


Pmarek12:
Hi @m-paz, @AdamSulek,

I have noticed that if we append data to .csv using CloudForCustomersReportToADLS flow, title columns are getting duplicated (horizontally)
![image](https://user-images.githubusercontent.com/48840222/143605725-54483882-4bc4-4def-9893-47f5361fdfb2.png)

AdamSulek:
- [ ] solved column duplication in "append" mode by adding sep parameter in read_csv

AdamSulek:
Here is PR for this issue: https://github.com/dyvenia/viadot/pull/220

PR #220, opened 2021-12-06
C4c adam

AdamSulek:
## Summary

Solving issue #209 

Adding tests for c4c_tasks 

## Checklist

This PR:


viadot/sources/cloud_for_customers.py
```diff
@@ -26,23 +23,24 @@ def __init__(
...
         Fetches data from Cloud for Customer.
-
         Args:
             report_url (str, optional): The url to the API in case of prepared report. Defaults to None.
             url (str, optional): The url to the API. Defaults to None.
             endpoint (str, optional): The endpoint of the API. Defaults to None.
             params (Dict[str, Any]): The query parameters like filter by creation date time. Defaults to json format.
             env (str, optional): The development environments. Defaults to 'QA'.
+            credentials (Dict[str, Any], optional): The credentials are provided from environmental variable
+            in local_config or with this parameter. Defaults to None than use credentials from local_config.
```
Trymzet, line 33: Config doesn't hold or populate environment variables, should be: `credentials are populated with values from config file or this parameter`.

viadot/tasks/cloud_for_customers.py
```diff
@@ -1,46 +1,132 @@
...
-    env: str = "QA",
-    if_empty: str = "warn",
-):
-    cloud_for_customers = CloudForCustomers(
-        url=url,
-        report_url=report_url,
-        endpoint=endpoint,
-        params=params,
-        env=env,
+class C4CReportToDF(Task):
```
Trymzet, line 8: Add docstrings on what these tasks do. Check out any Prefect task for an example.

tests/integration/tasks/test_cloud_for_customers.py
```diff
@@ -0,0 +1,24 @@
+from viadot.tasks import C4CToDF, C4CReportToDF
+from viadot.config import local_config
+
+
+def test_c4c_to_df():
+    credentials = local_config.get("CLOUD_FOR_CUSTOMERS")
+    credentials_prod = credentials["Prod"]
+    report_url = credentials_prod["server"]
```
Trymzet, line 8: if this is a report too, why do we have another task for reports? Otherwise, if it's not a report, it shouldn't be called `report_url`.

Trymzet (changes requested):
Last few changes

viadot/tasks/cloud_for_customers.py
```diff
@@ -1,46 +1,133 @@
...
+    def run(
+        self,
+        report_url: str = None,
+        env: str = "QA",
+        skip: int = 0,
+        top: int = 1000,
+    ):
+        """
+        Task for downloading data from the Cloud for Customers to a pandas DataFrame using report URL
+        (generated in Azure Data Factory).
```
Trymzet, line 49: Can you explain the difference between report and non-report here or point to C4C documentation explaining it?

viadot/tasks/bcp.py
```diff
@@ -95,6 +95,6 @@ def run(
         db_name = credentials["db_name"]
         uid = credentials["user"]
         pwd = credentials["password"]
-
+        print(f"server: {server}, db_name: {db_name}, uid: {uid}, pwd: {pwd}")
```
Trymzet, line 98: Revert this

viadot/tasks/azure_sql.py
```diff
@@ -99,6 +99,7 @@ def __init__(
         self.if_exists = if_exists
         self.credentials_secret = credentials_secret
         super().__init__(name="azure_sql_bulk_insert", *args, **kwargs)
+        print(f"credentials: {self.credentials_secret}")
```
Trymzet, line 102: Revert

viadot/sources/azure_sql.py
```diff
@@ -37,6 +37,7 @@ def bulk_insert(
         if schema is None:
             schema = self.DEFAULT_SCHEMA
         fqn = f"{schema}.{table}"
+        print(f"fqn: {fqn}")
```
Trymzet, line 40: Revert

viadot/sources/azure_sql.py
```diff
@@ -53,6 +54,7 @@ def bulk_insert(
         """
         if if_exists == "replace":
             self.run(f"DELETE FROM {schema}.{table}")
+        print(f"insert_sql: {insert_sql}")
```
Trymzet, line 57: Revert

viadot/flows/adls_to_azure_sql.py
```diff
@@ -245,6 +245,7 @@ def gen_flow(self) -> Flow:
         )
         bulk_insert_task.bind(
             path=self.local_file_path,
+            # from_path=self.local_file_path,
```
Trymzet, line 248: Revert

docker/run.sh
```diff
@@ -1,5 +1,5 @@
 IMAGE_TAG=latest
-PROFILE="user"
+PROFILE="dev"
```
Trymzet, line 2: Revert

viadot/sources/cloud_for_customers.py
```diff
@@ -84,19 +78,31 @@ def _to_records_report(self, url: str) -> List[Dict[str, Any]]:
...
+                tmp_full_url = url
+
             else:
                 # ODATA v1
                 new_records = response_json["d"]
+                url = None
+                self.params = None
+                self.endpoint = None
                 url = response_json.get("__next")
+                tmp_full_url = url
```
Trymzet, line 103: This whole bit is hard to understand. Why do we need `self.params` and `self.endpoint` to be set to `None`? And it's also repeated in two places? Also, there is no need to initialize `url` to `None` (eg. line 98) before assigning a value to it. It seems to me like you only need two lines for each case, the one with `new_records` and the one with `url`.

Trymzet (changes requested):

viadot/sources/cloud_for_customers.py
AdamSulek: Because we run in loop get_response() function which invokes connection, and in a second and higher number of invoking it does not need these parameters anymore. However, after ending these loop is good to set attributes to 'initial' values in terms of invoking to_df(), to_df() many times in jupyter notebook.

viadot/tasks/cloud_for_customers.py
AdamSulek: added

viadot/flows/adls_to_azure_sql.py
AdamSulek: revert

docker/run.sh
AdamSulek: revert

viadot/sources/azure_sql.py
AdamSulek: revert

viadot/tasks/azure_sql.py
AdamSulek: revert

viadot/tasks/bcp.py
AdamSulek: revert

Trymzet (approved):

PR #228, opened 2021-12-10
Delete all data from table

Rafalz13:
## Summary
a new option has been added which allows you to delete data from a table without deleting the table and creating it

## Importance
to avoid constantly renewing access to tables (which are removed with the "replace" option) for users

## Checklist

This PR:


tests/integration/test_azuresql.py
```diff
@@ -44,6 +49,13 @@ def test_create_table(azure_sql):
     assert table_object_id is not None
 
 
+def test_create_table_delete(azure_sql):
+    result = azure_sql.create_table(schema=SCHEMA, table=TABLE, if_exists="delete")
+    assert result == True
+    table_object_id = azure_sql.run(f"SELECT OBJECT_ID('{SCHEMA}.{TABLE}', 'U')")[0][0]
+    assert table_object_id is not None
```
Trymzet, line 56: This doesn't test what it's supposed to test. 

One way to test if the `delete` option works would be to have a table with some data, run the `if_exists="delete"`, and then verify that:
a) the table's creation date remained the same (sth like this is probably available in table metadata)
b) the data has been deleted

Trymzet (changes requested):
Test needs to be fixed (also, remove non-related commits)

PR #233, opened 2021-12-14
:sparkles: Add JSON to DataFrame task

lzuchowska:
This PR:

### DIFF
--- a/viadot/sources/azure_data_lake.py
+++ b/viadot/sources/azure_data_lake.py
@@ -157,19 +157,32 @@ class AzureDataLake(Source):
         path = path or self.path
         url = os.path.join(self.base_url, path)
 
-        if url.endswith(".csv"):
-            df = pd.read_csv(
-                url,
-                storage_options=self.storage_options,
-                sep=sep,
-                quoting=quoting,
-                lineterminator=lineterminator,
-                error_bad_lines=error_bad_lines,
-            )
-        elif url.endswith(".parquet"):
-            df = pd.read_parquet(url, storage_options=self.storage_options)
+        if os.path.splitext(os.path.split(path)[1])[1]:
+            if url.endswith(".csv"):
+                df = pd.read_csv(
+                    url,
+                    storage_options=self.storage_options,
+                    sep=sep,
+                    quoting=quoting,
+                    lineterminator=lineterminator,
+                    error_bad_lines=error_bad_lines,
+                )
+            elif url.endswith(".parquet"):
+                df = pd.read_parquet(url, storage_options=self.storage_options)
+            elif url.endswith(".json"):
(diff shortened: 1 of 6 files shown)


viadot/sources/azure_data_lake.py
```diff
@@ -157,19 +157,32 @@ def to_df(
...
+            elif url.endswith(".json"):
+                df = pd.read_json(url, storage_options=self.storage_options, lines=True)
+            else:
+                raise ValueError("Only CSV, parquet and JSON formats are supported.")
         else:
-            raise ValueError("Only CSV and parquet formats are supported.")
+            files = self.ls(path)
+            dfs = []
+            for file in files:
+                print(file)
```
Trymzet, line 180: To be removed

viadot/sources/azure_data_lake.py
```diff
@@ -157,19 +157,32 @@ def to_df(
...
+                df = pd.read_json(
+                    f"az://{file}", storage_options=self.storage_options, lines=True
+                )
```
Trymzet, line 183: Why do you assume that if it's a directory, then all the files inside it are JSONs and should be read with `pd.read_json()`? This should be checked by us, with the only assumption that all files in the directory should be of the same format (so we can check one file's extension and choose the loading method based on that). Eg.,
```
files = self.ls(path)
first_file_name, first_file_ext = os.path.splitext(files[0])
...
```
(decide based on extension to use `pd.read_csv()`, `pd.read_parquet()`, `dask_dataframe.read_json()`.

viadot/tasks/azure_data_lake.py
```diff
@@ -360,6 +360,32 @@ def run(
...
+
+    def run(self, dataframe: pd.DataFrame = None):
+        s = (dataframe.applymap(type) == list).all()
+        list_columns = s[s].index.tolist()
+
+        # applymap for columns that contains dicts (eg totals)
+        s = (dataframe.applymap(type) == dict).all()
+        dict_columns = s[s].index.tolist()
+
+        print(dict_columns)
```
Trymzet, line 380: To be removed

viadot/tasks/azure_data_lake.py
```diff
@@ -360,6 +360,32 @@ def run(
...
+
+        for col in dict_columns + list_columns:
+            print(col)
```
Trymzet, line 383: To be removed

viadot/tasks/azure_data_lake.py
```diff
@@ -360,6 +360,32 @@ def run(
         return df
 
 
+class AzureDataLakeFlattenDF(Task):
```
Trymzet, line 363: This task is in no way connected to Azure, should be just `FlattenDataFrame` and in `utils.py`.

Trymzet (changes requested):

PR #235, opened 2021-12-16
Add `delete` option to `Source.create_table()`'s `if_exists` parameter

Rafalz13:
## Summary
A new option has been added which allows you to delete data from a table without deleting the table and creating it.

## Importance
To avoid constantly renewing access to tables (which are removed with the "replace" option) for users.

## Checklist

This PR:

### DIFF
--- a/tests/integration/test_azuresql.py
+++ b/tests/integration/test_azuresql.py
@@ -49,14 +49,19 @@ def test_create_table_replace(azure_sql):
     assert table_object_id is not None
 
 
-def test_create_table_delete(azure_sql):
-    dtypes = {"country": "VARCHAR(100)", "sales": "FLOAT(24)"}
-    result_replace = azure_sql.create_table(
-        schema=SCHEMA, table=TABLE, dtypes=dtypes, if_exists="replace"
+def test_create_table_delete(azure_sql, TEST_CSV_FILE_BLOB_PATH):
+    executed = azure_sql.bulk_insert(
+        schema=SCHEMA,
+        table=TABLE,
+        source_path=TEST_CSV_FILE_BLOB_PATH,
+        if_exists="replace",
     )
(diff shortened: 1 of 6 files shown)


tests/integration/test_azuresql.py
```diff
@@ -44,6 +49,36 @@ def test_create_table(azure_sql):
...
+    assert int(result[0][0]) == 230
+
+    table_object_id_insert = azure_sql.run(
+        f"SELECT OBJECT_ID('{SCHEMA}.{TABLE}', 'U')"
+    )[0][0]
+
+    result_delete = azure_sql.create_table(
+        schema=SCHEMA, table=TABLE, if_exists="delete"
+    )
+    assert result_delete == True
```
Trymzet, line 71: Should be `is True`, like in line 59. Also here you're only testing if the delete statement was executed, so it could use the same naming as above where you're testing if the insert statement was executed (`delete_executed`, while changing `executed` to `insert_executed`).

viadot/sources/sharepoint.py
```diff
@@ -52,7 +52,8 @@ def download_file(
         download_to_path: str = "Sharepoint_file.xlsm",
     ) -> None:
 
-        if not download_from_path or self.download_from_path:
+        download_from_path = download_from_path or self.download_from_path
```
Trymzet, line 55: If you want to keep the path as an attribute (so it can be used in eg. self.to_df()), I would call it just `path` instead (as it can be used for other things than downloading). So, in `download_file()`, you would have: `download_from_path = download_from_path or self.path`. Otherwise, deleting this attribute completely would also be fine since it's only used in `download_file()`.

Trymzet (changes requested):
Looks good, minor changes

Trymzet (approved):

PR #246, opened 2022-01-13
Added task to check column order to ADLSToAzureSQL Flow

Rafalz13:
## Summary
Added task to check column order to ADLSToAzureSQL Flow. If the order is different in the loaded DataFrame and in the SQL table where data from the CSV file are loaded flow is stopped.

## Importance
To improve the quality of data in the database and improve data validation.

## Checklist

This PR:


viadot/flows/adls_to_azure_sql.py
```diff
@@ -74,6 +78,28 @@ def df_to_csv_task(df, path: str, sep: str = "\t"):
     df.to_csv(path, sep=sep, index=False)
 
 
+@task
+def check_column_order_task(
+    table: str = None,
+    df: DataFrame = None,
+    sqldb_credentials_secret: str = None,
+    if_exists: str = None,
```
lzuchowska, line 86: Let's keep the convention the same as in constructor parameters ✨ 

```suggestion
   if_exists: Literal["fail", "replace", "append", "delete"] = "replace",
```

However, add only those values that are being handled by the method.

lzuchowska (commented):
It might be better to place this functionality inside of `task_utils.py`, what do you think?

Trymzet (changes requested):

viadot/tasks/azure_sql.py
```diff
@@ -264,3 +267,72 @@ def run(
...
+        self.vault_name = vault_name
+
+        super().__init__(name="run_check_column_order", *args, **kwargs)
+
+    def df_change_order(
+        self, df: pd.DataFrame = None, sql_column_list: List[str] = None
+    ):
+        df_column_list = list(df.columns)
+        if set(df_column_list) == set(sql_column_list):
+            df_column_list = sql_column_list
```
lzuchowska, line 298: This line doesn't really change anything, as `df_column_list` is not used anywhere. Was it meant to be used later in `df_changed`?

In line 331 we are checking the equality of DataFrame columns and SQL columns. Is it really necessary to check for it again?

lzuchowska (changes requested):
Tests are passing ✔️  I think it will be ready to merge once this one problem is solved.

viadot/tasks/azure_sql.py
Rafalz13: A new version of the file was uploaded.
Line 334 is a comparison of two lists that are identical (whether the order is the same).
Line 297  is a comparison of the number of columns if they are the same (can be in a different order) - if so - the order is changed if not - an exception is thrown.
Make sense?

lzuchowska: Yup, I see. Looks fine to me 😄

lzuchowska (approved):
Looks good to me 👍

PR #390, opened 2022-04-27
SAP RFC To ADLS

angelika233:
## Summary
Added parameter func in SAPRFCToDF task and SAPRFC source and added new flow SAPRFCToADLS

## Importance
Added a new flow and extra parameter

## Checklist

This PR:


viadot/flows/sap_rfc_to_adls.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class SAPRFCToADLS(Flow):
+    def __init__(
+        self,
+        name: str,
+        query_list: List[str],
+        sep: str = None,
+        func: str = "BBP_RFC_READ_TABLE",
+        sap_credentials: dict = None,
```
Trymzet, line 36: This should be either an Azure KV or Prefect secret, but we can agree this will be added in a follow up PR

viadot/flows/sap_rfc_to_adls.py
```diff
@@ -0,0 +1,118 @@
...
+        local_file_path: str = None,
+        file_sep: str = "\t",
+        if_exists: Literal["append", "replace", "skip"] = "replace",
+        adls_path: str = None,
+        overwrite: bool = False,
+        sp_credentials_secret: str = None,
```
Trymzet, line 42: This should specify `adls` since we will eventually use two credential secrets, one for SAP and one for ADLS. So I would call it eg. `adls_sp_credentials_secret`, and then later on we will have `sap_credentials_secret`.

viadot/flows/sap_rfc_to_adls.py
```diff
@@ -0,0 +1,118 @@
...
+        Unsupported:
+            - aggregations
+            - joins
+            - subqueries
+            - etc.
+
+        Args:
+            name (str): The name of the flow.
+            query_list(List[str]) The list of queries to be executed with pyRFC.
+            sep(str, optional): Which separator to use when querying SAP. If not provided, multiple options are automatically tried.
```
Trymzet, line 64: I think if you qualify one, you should qualify the other, so if you have `file_sep`, this should be eg. `query_sep`, `rfc_sep`, or similar.

viadot/flows/sap_rfc_to_adls.py
```diff
@@ -0,0 +1,118 @@
...
+            flow=self,
+        )
+        df_full = concat_dfs.bind(df, flow=self)
+        csv = df_to_csv.bind(
+            df=df_full,
+            sep=self.file_sep,
+            path=self.local_file_path,
+            if_exists=self.if_exists,
+            flow=self,
+        )
```
Trymzet, line 108: We load to the lake in parquet by default, if we need to load to a CSV this should be handled by a parameter. See `SupermetricsToADLS` for reference.

Trymzet (changes requested):
Great work! I added my comments and suggestions.

tests/integration/flows/test_sap_rfc_to_adls.py
```diff
@@ -0,0 +1,25 @@
...
+        func="BBP_RFC_READ_TABLE",
+        sap_credentials=sap_test_creds,
+        local_file_path="test_file.csv",
+        adls_path=ADLS_PATH,
+        overwrite=True,
+    )
+    result = flow.run()
+    assert result.is_successful()
+    file = AzureDataLake(ADLS_PATH)
+    assert file.exists()
```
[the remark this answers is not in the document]
angelika233: I'm not sure, maybe it would be more clear to specify credentials but some other tests (for example supermetrics to ADLS) don't specify it and I wanted to be consistent

viadot/task_utils.py
```diff
@@ -462,6 +462,22 @@ def df_clean_column(
...
+
+    Args:
+        dfs (List[pd.DataFrame]): List of dataframes to concat.
+    Returns:
+        full_df (pd.DataFrame()): Pandas dataframe containing all columns from dataframes from list.
+    """
+    full_df = pd.DataFrame()
+    for df in dfs:
+        full_df = pd.concat([full_df, df], axis=1)
+    return full_df
```
Trymzet: You forgot to implement my comment, the standard way to do it is to do `pd.contact[dfs]`, or is there any specific reason to do it in a loop?

viadot/flows/sap_rfc_to_adls.py
```diff
@@ -4,48 +4,34 @@
...
 class SAPRFCToADLS(Flow):
     def __init__(
         self,
         name: str,
-        query_list: List[str],
-        sep: str = None,
+        query: str = None,
+        queries: List[str] = None,
+        rfc_sep: str = None,
         func: str = "BBP_RFC_READ_TABLE",
```
Trymzet: default `func` should be consistent with the default in the source (so either set to the same or change in both places if appropriate

Trymzet (changes requested):
Only two last things

viadot/task_utils.py
angelika233: It's already changed, you can see it in the "Files changed". I don't know why here is an old version.

viadot/flows/sap_rfc_to_adls.py
angelika233: same as above :)

Trymzet (approved):

PR #434, opened 2022-06-15
✨Added databricks source to viadot

afraijat:
## Summary

Added databricks source to viadot library

## Importance

Added databricks source to viadot library with the following functionalities:
1) Creating, reading and deleting tables
2) Converting to and from Spark DataFrames
3) Create and delete schemas
4) Upserting records
5) Full-refresh
6) Printing data schema of a table

## Checklist

This PR:


viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
+import json
+from pandas import DataFrame
```
Trymzet, line 2: This import is confusing as you're dealing with two types of DataFrames in this file. This should be imported as eg. `pd.DataFrame` to always know which DataFrame class we're dealing with.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+    def connect(self):
+        if self.env == "QA":
+            self.spark = SparkSession.builder.getOrCreate()
+        else:
+            self.spark = self._get_spark_session(self.env)
+    
+    def to_df(self, list):
+        return pd.DataFrame(list)
+    
+    def to_json(sparkDf: pyspark.sql.dataframe.DataFrame):
```
Trymzet, line 73: I would import this as `spark.DataFrame` for clarity. Then you have `pd.DataFrame` and `spark.DataFrame`.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
+import json
+from pandas import DataFrame
+from viadot.config import local_config
+from viadot.sources.base import Source
+from delta.tables import *
+import pandas as pd
+import pyspark
+from pyspark import SparkConf, SparkContext
+import time
+# from viadot.sources import Base
```
Trymzet, line 10: Remove

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+class Databricks(Source):
```
Trymzet, line 11: Use Black to format the code; there should be two spaces before a class definition.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        ```
+        Returns:
+            pyspark.sql.dataframe.DataFrame: The converted Spark DataFrame
+        """
+        sparkDF = self.spark.createDataFrame(df)
+        return sparkDF
+    
+    def spark_to_pandas(self, sparkDf: pyspark.sql.dataframe.DataFrame):
+        """
+        Convert Spark DataFrame to Pandas DataFrame
```
Trymzet, line 98: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        list = [{"some":"data"}]
+        sparkDf = databricks.to_spark_df(df)
+        pandasDf = spark_to_pandas(sparkDf)
+        ```
+        Returns:
+            pyspark.sql.dataframe.DataFrame: The converted Spark DataFrame
```
Trymzet, line 112: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        return sparkDf.toPandas()
+    
+    def query(self, query: str):
+        """
+        Execute an SQL query
```
Trymzet, line 118: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+            query (str): An SQL query
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        queryResult = databricks.query("SELECT * FROM schema.table_1")
+        ```
+        Returns:
+            DataFrame: A Pandas DataFrame containing the query's results (if available)
```
Trymzet, line 130: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        result = self.spark.sql(query)
+        return self.spark_to_pandas(result)
+    
+    def read_table(self, table_name: str):
+        """
+        Read the contents of a given table
```
Trymzet, line 137: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            table_name (str): Name of an existing table
```
Trymzet, line 140: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        tableData = databricks.read_table("schema.table_1")
+        ```
+        Returns:
+            DataFrame: A Pandas DataFrame containing the requested table's data
```
Trymzet, line 149: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        result = self.spark.sql("select * from "+ table_name)
+        return self.spark_to_pandas(result)
+
+    def create_table(self, table_name: str, df: DataFrame):
+        """
+        Write a new table using a given Pandas DataFrame
```
Trymzet, line 156: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            table_name (str): Name of the new table to be created
+            df (DataFrame): DataFrame to be written as a table
```
Trymzet, line 160: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        df = [{"id":"1", "name":"Joe"}]
+        new_table = databricks.create_table("schema.table_1", df)
+        ```
+        Returns:
+            DataFrame: A Pandas DataFrame containing the newly created table's data
```
Trymzet, line 170: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        data.createOrReplaceTempView("new_table")
+
+        self.query("CREATE TABLE " + table_name +" USING DELTA AS SELECT * FROM new_table;")
+
+        print("Table " +table_name+ " created.")
+        return self.read_table(table_name)    
+
+    def delete_table(self, table_name: str):
+        """
+        Delete an existing table
```
Trymzet, line 182: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            table_name (str): Name of the table to be overwritten
+            df (DataFrame): DataFrame to be used to overwrite the table
```
Trymzet, line 186: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        databricks = Databricks()]
+        databricks.delete_table("schema.table_1")
+        ```
+        """
+        self.spark.sql("DROP TABLE "+ table_name)
+        print("Table deleted.")
+
+    def full_refresh(self, table_name: str, df: DataFrame):
+        """
+        Overwrite an existing table with data from a Pandas DataFrame
```
Trymzet, line 200: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            table_name (str): Name of the table to be overwritten
+            df (DataFrame): DataFrame to be used to overwrite the table
```
Trymzet, line 204: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        df = [{"id":"1", "name":"Joe"}]
+        new_table = databricks.full_refresh("schema.table_1", df)
+        ```
+        Returns:
+            DataFrame: A Pandas DataFrame containing the new data in the overwitten table
```
Trymzet, line 214: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        data = self.to_spark_df(df)
+        data.write.format("delta").mode("overwrite").saveAsTable(table_name)
+        return self.read_table(table_name)
+        
+    # TODO: add a method for extracting a usable JSON schema for the sets
+    # TODO: Parse the matchedSet as a notMatchedSet
+    def upsert(self, table_name: str, df: DataFrame, matchedSet: dict, notMatchedSet: dict, pk: str):
+        """
+        Upsert the data to an existing table. If the record exists, the function updates the record in the table.
+        If it doesn't exist, the function creates a new record in the table
```
Trymzet, line 225: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            table_name (str): Name of the table to perform the upsert operation on
+            df (DataFrame): DataFrame to be used to upsert the table
+            matchedSet (dict): A dictionary defining the changes that should take place in case a matching record is found
+            notMatchedSet(dict): A dictionary defining the fields that will be added in case no records exist
+            pk (str): The primary key that should be joined on
```
Trymzet, line 232: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+    def upsert(self, table_name: str, df: DataFrame, matchedSet: dict, notMatchedSet: dict, pk: str):
+        """
+        Upsert the data to an existing table. If the record exists, the function updates the record in the table.
+        If it doesn't exist, the function creates a new record in the table
+
+        Args:
+            table_name (str): Name of the table to perform the upsert operation on
+            df (DataFrame): DataFrame to be used to upsert the table
+            matchedSet (dict): A dictionary defining the changes that should take place in case a matching record is found
+            notMatchedSet(dict): A dictionary defining the fields that will be added in case no records exist
```
Trymzet, line 231: Remove camelcase

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        data = [{"Id": "KVSzUaILfQZXDb", "wrong": "Updated3!EHNYKjSZsiy", "Name": "Upsert-Black", "FirstName": "Updated", "LastName": "Carter2", "ContactEMail": "Adam.Carter@TurnerBlack.com", "MailingCity": "Updated!Jamesport"}]
+        match = {"Id": "main.Id",
+                "AccountId": "updates.AccountId",
+                "Name": "updates.Name",
+                "FirstName": "updates.FirstName",
+                "LastName": "updates.LastName",
+                "ContactEMail": "updates.ContactEMail",
+                "MailingCity": "main.MailingCity"}
+
+        notMatch = {
```
Trymzet, line 246: remove camelcase

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+            " WHEN MATCHED THEN UPDATE SET " + matchedString + " WHEN NOT MATCHED THEN INSERT (" + notMatchedKeys
+            + ") VALUES (" + notMatchedValues + ")")
+
+            print("Data upserted successfully.")
+        else:
+            print("Columns mismatch. Please adhere to the column names and data types used in the source table.")
+    
+    def create_schema(self, schemaName: str):
+        """
+        Create a schema for storing tables
```
Trymzet, line 302: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            schemaName (str): Name of the new schema to be created
```
Trymzet, line 305: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        databricks = Databricks()
+        databricks.create_schema("schema_1")
+        ```
+        """
+        self.spark.sql("CREATE SCHEMA IF NOT EXISTS " + schemaName)
+        print("Schema "+ schemaName + " created.")
+
+    def delete_schema(self, schemaName: str):
+        """
+        Delete a schema
```
Trymzet, line 319: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            schemaName (str): Name of the schema to be deleted
```
Trymzet, line 322: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        databricks = Databricks()
+        databricks.delete_schema("schema_1")
+        ```
+        """
+        self.spark.sql("DROP SCHEMA " + schemaName)
+        print("Schema "+ schemaName + " deleted.")
+
+    def print_schema(self, table_name: str):
+        """
+        Print the details of a table's data schema
```
Trymzet, line 336: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            table_name (str): Name of the table whose schema will be printed
```
Trymzet, line 339: Use dots at the end of sentences

tests/integration/tasks/test_databricks.py
```diff
@@ -0,0 +1,98 @@
...
+    table_name = "raw.test_table"
+    match = {"Id": "main.Id",
+            "AccountId": "updates.AccountId",
+            "Name": "updates.Name",
+            "FirstName": "updates.FirstName",
+            "LastName": "updates.LastName",
+            "ContactEMail": "updates.ContactEMail",
+            "MailingCity": "main.MailingCity"}
+
+    notMatch = {
```
Trymzet, line 82: Remove camelcase

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+    
+    def to_df(self, list):
+        return pd.DataFrame(list)
+    
+    def to_json(sparkDf: pyspark.sql.dataframe.DataFrame):
+        return sparkDf.toJSON()
+    
+    def to_spark_df(self, df: DataFrame):
+        """
+        Convert Pandas DataFrame to Spark DataFrame
```
Trymzet, line 78: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+
+        Args:
+            df (DataFrame): Pandas DataFrame to be converted to Spark DataFrame.
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        list = [{"some":"data"}]
+        sparkDf = databricks.to_spark_df(list)
```
Trymzet, line 88: The example will have to be adjusted once the method is adjusted as per comments.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        ```
+        Returns:
+            pyspark.sql.dataframe.DataFrame: The converted Spark DataFrame
+        """
+        sparkDF = self.spark.createDataFrame(df)
+        return sparkDF
```
Trymzet, line 94: Remove camel case

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+            pyspark.sql.dataframe.DataFrame: The converted Spark DataFrame
+        """
+        return sparkDf.toPandas()
+    
+    def query(self, query: str):
+        """
+        Execute an SQL query
+
+        Args:
+            query (str): An SQL query
```
Trymzet, line 121: Use dots at the end of sentences

viadot/sources/databricks.py
```diff
@@ -0,0 +1,350 @@
...
+        databricks = Databricks()
+        queryResult = databricks.query("SELECT * FROM schema.table_1")
+        ```
+        Returns:
+            DataFrame: A Pandas DataFrame containing the query's results (if available)
+        """
+        result = self.spark.sql(query)
+        return self.spark_to_pandas(result)
+    
+    def read_table(self, table_name: str):
```
Trymzet, line 135: Not sure if we need that if we have `query()` and `to_df()` already?

Trymzet (changes requested):
I'm probably not finished but I need the code to be a bit clearer to review further; there's plenty to work on anyway.

viadot/sources/databricks.py
```diff
@@ -44,146 +57,160 @@ def _get_spark_session(self, env: str = None):
...
+            str: A JSON string containing the requested table's data.
         """
-        return sparkDf.toPandas()
-    
-    def query(self, query: str):
+        result = self.sc.sql("select * from " + table_name)
+        df = self._spark_df_to_pandas_df(result)
+        if df.empty:
+            self._handle_if_empty(if_empty)
+        return df.to_json()
```
Trymzet, line 108: The purpose of `to_json()` is to offer a more flexible way to consume data (not necessarily a pandas DataFrame). If you need to first load it to a DataFrame anyway then this function doesn't make sense -- user can just use `Databricks.to_df()`.

viadot/sources/databricks.py
```diff
@@ -44,146 +57,160 @@ def _get_spark_session(self, env: str = None):
...
 
         Example:
         ```python
         from viadot.sources import Databricks
         databricks = Databricks()
-        tableData = databricks.read_table("schema.table_1")
+        list = [{"id":"1", "name":"Joe"}]
+        df = pd.DataFrame(list)
+        spark_df = databricks._pandas_df_to_spark_df(df)
+        pandas_df = _spark_df_to_pandas_df(spark_df)
```
Trymzet, line 146: Should be: `databricks._spark_df_to_pandas_df(spark_df)`

viadot/sources/databricks.py
```diff
@@ -44,146 +57,160 @@ def _get_spark_session(self, env: str = None):
...
+            spark.DataFrame: The converted Spark DataFrame.
         """
-        result = self.spark.sql(query)
-        return self.spark_to_pandas(result)
-    
-    def read_table(self, table_name: str):
+        spark_df = self.sc.createDataFrame(df)
+        return spark_df
+
+    def _spark_df_to_pandas_df(self, spark_df: spark.DataFrame):
```
Trymzet, line 132: Missing return type annotation.

Trymzet (changes requested):

tests/integration/tasks/test_databricks.py
```diff
@@ -305,35 +305,40 @@ def test_upsert():
...
-    match = {
-        "Id": "main.Id",
-        "AccountId": "updates.AccountId",
-        "Name": "updates.Name",
-        "FirstName": "updates.FirstName",
-        "LastName": "updates.LastName",
-        "ContactEMail": "updates.ContactEMail",
-        "MailingCity": "main.MailingCity",
-    }
+    match = "Id= main.Id, AccountId= updates.AccountId, Name= updates.Name, FirstName= updates.FirstName, LastName= updates.LastName, ContactEMail= updates.ContactEMail, MailingCity= main.MailingCity"
```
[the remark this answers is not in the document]
afraijat: 1. Noted and fixed

2. Previous implementations used dictionaries, and the final result needs to be a string. Both got rejected (first because the dictionaries were converted to a string, second because it was different from `utils.build_merge_query()`) and as such, I opted to using a string directly.

3. `utils.build_merge_query()` assumes that the user wants to update all the fields in every single upsert operation; it does not give them the option of selectively updating certain fields while allowing other fields to retain their old values. This function allows them to specify which fields will be updated, and which ones will be left intact. For example:
` match="Id=main.Id, AccountId=updates.AccountId, Name=updates.Name, FirstName=updates.FirstName, LastName=updates.LastName, ContactEMail=updates.ContactEMail, MailingCity=main.MailingCity"`
Here, we specify that we want to insert the value of `MailingCity` corresponding to the `main` table, which is the existing/main table that we read from Databricks; in other words, this field won't be updated. The other columns (besides the primary key, but that's irrelevant to the end result) have their values updated using the values from the `updates` temporary view, which is the staging table.

4. For a. the reason described in point 3 and b. The query used to retrieve the column names in `utils.build_merge_query()` does not work with Databricks/SparkSQL; it requires a different SQL statement

5. Noted and fixed

tests/integration/tasks/test_databricks.py
```diff
@@ -305,35 +305,40 @@ def test_upsert():
...
+    data = pd.DataFrame(data)
+    query = databricks.build_merge_query(table_name, "Id", data, match)
 
     df = pd.DataFrame(source_data)
     df = df.sort_values(by="Id")
     df = df.reset_index(drop=True)
 
-    databricks.upsert(table_name, upsert_data, match, "Id")
+    databricks.query(query)
```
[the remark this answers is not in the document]
afraijat: 1. `databricks.upsert()` was using `databricks.query()` before to run the upsert statement internally; the the difference is now the query building and execution are split up.

2. `databricks.build_merge_query()` builds the upsert query only; it does not run it. We need to run the query/statement somehow. If the name `query` is confusing, I can rename the `query` function to `sql`; this way it would be clearer that it's for general SQL statements, both querying data and performing operations (ex. create table/schema, delete table etc.)

viadot/sources/databricks.py
```diff
@@ -245,82 +245,48 @@ def full_refresh(self, table_name: str, df: pd.DataFrame):
...
-                + " VALUES "
-                + columns
-            )
-
-            print("Data upserted successfully.")
-        else:
-            print(
-                "Columns mismatch. Please adhere to the column names and data types used in the source table."
-            )
+        upsertData.createOrReplaceTempView("updates")
```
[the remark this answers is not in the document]
afraijat: As mentioned in the other comment, the function in `utils.py` uses a different SQL statement for retrieving the column names and does not support selectively updating fields of a record; it cannot be reused without changing its implementation

Trymzet (changes requested):

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def run(
+        self, query: str, fetch_type: Literal["record", "dataframe"] = "record"
+    ) -> Union[spark.DataFrame, pd.DataFrame, bool]:
+        """
+        Execute an SQL query.
+
+        Args:
+            query (str): The query to execute.
+            fetch_type (Literal[, optional): How to return the data: either
```
Trymzet, line 146: Typo in the typing

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+            schema_name (str): Name of the new schema to be created.
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        databricks.create_schema("schema_1")
+        ```
+        """
+        self.session.sql(f"CREATE SCHEMA IF NOT EXISTS {schema_name}")
```
Trymzet, line 348: Should use `self.run()`

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+            schema_name (str): Name of the schema to be deleted.
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        databricks.delete_schema("schema_1")
+        ```
+        """
+        self.session.sql(f"DROP SCHEMA {schema_name}")
```
Trymzet, line 365: Should use `self.run()`

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+        databricks = Databricks()
+        databricks.discover_schema("table_1")
+        ```
+        Returns:
+            schema: A dictionary containing the schema details of the table.
+        """
+        result = self.run(f"DESCRIBE {table_name}", "dataframe")
+
+        # Delete the last 3 records, which are boilerplate for Delta Tables
+        names = result["col_name"].values.tolist()
```
Trymzet, line 388: I think `col_names` or `column_names` would be a more precise name.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+        del names[-3:]
+
+        data_types = result["data_type"].values.tolist()
+        del data_types[-3:]
```
Trymzet, line 392: In Python it's preferred to slice rather than modify arrays. You can do `names = names[:-2]`.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+
+        schema = {names[i]: data_types[i] for i in range(len(names))}
```
Trymzet, line 394: It seems like you could use `zip()` here (I think `dict(zip(col_names, dtypes))` should work).

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+        databricks.discover_schema("table_1")
+        ```
+        Returns:
+            schema: A dictionary containing the schema details of the table.
+        """
+        result = self.run(f"DESCRIBE {table_name}", "dataframe")
```
Trymzet, line 385: It seems like you make your life harder by fetching this is a DataFrame, can't we fetch this as a list or dict?

viadot/sources/databricks.py
```diff
@@ -0,0 +1,396 @@
...
+            result = True
+
+        return result
+
+    def _check_if_table_exists(self, table_name: str) -> bool:
+        split_name = table_name.split(".")
+        does_exist = self.session._jsparkSession.catalog().tableExists(
+            split_name[0], split_name[1]
+        )
+        return does_exist
```
Trymzet, line 186: Refer back to the comment about splitting table and schema as separate parameters.

Trymzet (changes requested):

viadot/sources/databricks.py
afraijat: Unfortunately no, since it's a Spark sql statement running, it's not possible to directly fetch it as a list or dict.

Trymzet: Then why do we have the `fetch_type` parameter at all? 😅 If you don't use it then you can remove it. Then this stays the same just without having to pass "dataframe" as the argument to `fetch_type`.

afraijat: I'm not sure what you mean. By default, it fetches it as a Spark DataFrame (the default record type), but I want it to be returned as a Pandas DataFrame for the ease of manipulating it.

Trymzet:
I mean if we never need anything else than a pandas dataframe then you can
remove the fetch type parameter.

Trymzet:
If we do use the Spark dataframe as well then you're right we need to
specifiy the return type.

However then pls change the name as you have "dataframe" which is the most
confusing name possible since you literally can only return two dataframe
types (spark and pandas).

Trymzet:
Pls also add .prefect to .gitignore as discussed

afraijat:
> 

I checked, `.prefect` is already in `.gitignore`. I'm not sure how the folder got pushed.

tests/integration/tasks/test_databricks.py
```diff
@@ -0,0 +1,183 @@
...
+            "FirstName": "Melody",
+            "LastName": "Cook",
+            "ContactEMail": "Melody.Cook@ScottMerritt.com",
+            "MailingCity": "Elizabethfurt",
+        }
+    ]
+
+    full_refresh_df = pd.DataFrame(full_refresh_data)
+
+    did_func_work = databricks.insert_into(
```
lzuchowska, line 116: Same case as the previous comment.

tests/integration/tasks/test_databricks.py
```diff
@@ -0,0 +1,183 @@
...
+
+    table = "test_upsert"
+    fqn = f"{schema}.{table}"
+    primary_key = "Id"
+
+    databricks.create_table_from_pandas(schema, table, df)
+
+    upsert_df = pd.DataFrame(upsert_data)
+
+    did_func_work = databricks.insert_into(
```
lzuchowska, line 151: Same as the previous comment

viadot/sources/databricks.py
```diff
@@ -0,0 +1,436 @@
...
+    def __init__(
+        self, env: str = "DEV", credentials: Dict[str, Any] = None, *args, **kwargs
+    ):
+        self.env = env
+
+        self.credentials = credentials or local_config.get("DATABRICKS", {}).get(env)
+
+        self.connect()
+
+    def _get_spark_session(self, env: str = "DEV"):
```
lzuchowska, line 52: This method created a new spark session, reads config, adjusts it and returns a `new_spark`. Shouldn't it be named more like `_create_spark_session` than `_get_spark_session`?

Moreover, we could name `new_config` just as `config`, as there is no *old* config or any other local config variable.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,436 @@
...
+                f"Only the values {allowed_fetch_type_values} are allowed for 'fetch_type'"
+            )
+
+        query_clean = query.upper().strip()
+        query_keywords = ["SELECT", "SHOW", "PRAGMA", "DESCRIBE"]
+
+        query_result = self.session.sql(query)
+
+        if any(query_clean.startswith(word) for word in query_keywords):
+            if fetch_type == "record":
```
lzuchowska, line 194: Correct me if I'm wrong but possible values for `fetch_type` are `spark` and `pandas` as far as I can see, so this statement would never execute.

viadot/sources/databricks.py
```diff
@@ -0,0 +1,436 @@
...
+        Delete an existing table.
+
+        Args:
+            schema (str): Name of the schema
+            table (str): Name of the new table to be created.
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()]
```
lzuchowska, line 255: Please fix a typo at `Databricks()]`

viadot/sources/databricks.py
```diff
@@ -0,0 +1,436 @@
...
+    ):
+        """
+        Insert data into a table.
+
+        Args:
+            schema (str): Name of the schema
+            table (str): Name of the new table to be created.
+            df (pd.DataFrame): DataFrame with the data to be inserted into the table.
+            primary_key (str, Optional): Needed only when updating the data, the primary key on which the data will be joined.
+            if_exists (str, Optional): Which operation to run with the data. Allowed operations are: replace, append and update.
```
lzuchowska, line 344: `if_exists` can also have a value `fail`, please correct that

lzuchowska (changes requested):
To summarize:

- [ ] `did_func_work` naming change in line 80. in `test_databricks.py`
- [ ]  `did_func_work` naming change in line 116. in `test_databricks.py`
- [ ]  `did_func_work` naming change in line 151. in `test_databricks.py`
- [ ] please add some tests for negative cases - when insert *shouldn't* happen, make sure those cases are covered up too.
- [ ]  name change in `databricks.py` in line 69.
- [ ] rename method name and variables inside `_get_spark_session()`
- [ ] `if fetch_type == "record"` statement correction in `run` method
- [ ] fix typo in docstrings in `drop_table()` method
- [ ] change the `fetch_type` check in `run()` method
- [ ] fix docstrings in `insert_into()` method
- [ ] change ifs in `insert_into()` method
- [ ] Remove comment above `discover_schema()` method

lzuchowska:
Looks fine to me!

lzuchowska (approved):

PR #435, opened 2022-06-17
Empty parquet handling

angelika233:
## Summary
Added `if_empty` parameter in `DuckDBCreateTableFromParquet` task and in `EpicorToDuckDB`, `SAPToDuckDB`,
`SQLServerToDuckDB` . Added `check_if_empty_file()` and `handle_if_empty_file()` in `utils.py`

Added `map_dtypes_to_str` task in `utils.py` and added this to `EpicorToDuckDB`, `SAPToDuckDB`, `SQLServerToDuckDB`

## Importance
When Parquet was empty (or had only one column `_viadot_downloaded_at_utc`) DuckDB couldn't upload file to DB because of column mismatch. Now there is an option to skip if file is empty

Added mapping to strings  before uploading data to DuckDB

## Checklist

This PR:

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,5 +1,10 @@
-from viadot.utils import gen_bulk_insert_query_from_df
 import pandas as pd
+import pytest
+import logging
+import os
+
+from viadot.utils import gen_bulk_insert_query_from_df, check_if_empty_file
+from viadot.signals import SKIP
 
 
 def test_single_quotes_inside():
@@ -66,3 +71,33 @@ def test_double_quotes_inside():
 
 VALUES ({TEST_VALUE_ESCAPED}, 'c')"""
     ), test_insert_query
+
+
+def test_check_if_empty_file(caplog):
+    with open("empty.csv", "w"):
+        pass
+
+    with caplog.at_level(logging.WARNING):
+        check_if_empty_file(path="empty.csv", if_empty="warn", file_extension=".csv")
+        assert "The input file is empty." in caplog.text
+    with pytest.raises(ValueError):
+        check_if_empty_file(path="empty.csv", if_empty="fail", file_extension=".csv")
+    with pytest.raises(SKIP):
+        check_if_empty_file(path="empty.csv", if_empty="skip", file_extension=".csv")
+
+    os.remove("empty.csv")
+
+
+def test_check_if_empty_file_one_column(caplog):
+    df = pd.DataFrame(columns=["_viadot_downloaded_at_utc"], index=[0])
+    df.to_csv("empty.csv", index=False)
+
+    with caplog.at_level(logging.WARNING):
+        check_if_empty_file(path="empty.csv", if_empty="warn", file_extension=".csv")
+        assert "The input file is empty." in caplog.text
+    with pytest.raises(ValueError):
+        check_if_empty_file(path="empty.csv", if_empty="fail", file_extension=".csv")
+    with pytest.raises(SKIP):
+        check_if_empty_file(path="empty.csv", if_empty="skip", file_extension=".csv")
+
+    os.remove("empty.csv")
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -6,6 +6,9 @@ import os
 from viadot.utils import gen_bulk_insert_query_from_df, check_if_empty_file
 from viadot.signals import SKIP
 
+EMPTY_CSV_PATH = "empty.csv"
+EMPTY_PARQUET_PATH = "empty.parquet"
+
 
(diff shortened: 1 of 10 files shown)


viadot/task_utils.py
```diff
@@ -475,6 +475,23 @@ def concat_dfs(dfs: List[pd.DataFrame]):
...
+    Task for mapping dtypes to strings.
+
+    Args:
+        df (pd.DataFrame): input DataFrame.
+
+    Returns:
+        df_mapped (pd.DataFrame): Pandas DataFrame with mapped Data Types.
+    """
+    df_mapped = df.copy()
+    for col in df_mapped.columns:
```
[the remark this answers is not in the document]
Trymzet: I think if you're casting the whole df you can just cast at df level (eg. `df.astype(str)`), not column by column. But why are you casting the whole df to string in the first place? It's highly inefficient from storage and compute perspective, so you should properly explain in the code why you're doing it.

Also `Task for mapping dtypes to strings.` -- this is not true, dtypes would most likely be a dict and the input to this function is a pandas DataFrame. What you're doing is casting an entire DataFrame to a string data type.

viadot/utils.py
```diff
@@ -345,3 +351,31 @@ def union_dict(*dicts):
...
+def check_if_empty_file(
+    path: str,
+    if_empty: Literal["warn", "skip", "fail"] = "warn",
+    file_extension: Literal[".parquet", ".csv"] = None,
+    file_sep: str = "\t",
+):
+    if os.stat(path).st_size == 0:
+        handle_if_empty_file(if_empty)
+
+    elif file_extension == ".parquet":
```
Trymzet, line 374: Should include a comment why you're doing this (why not use `os.stat` and why use `_viadot_downloaded_at_utc` column)

viadot/utils.py
```diff
@@ -345,3 +351,31 @@ def union_dict(*dicts):
...
+        df = pd.read_parquet(path)
+        if "_viadot_downloaded_at_utc" in df.columns and len(df.columns) == 1:
+            handle_if_empty_file(if_empty)
+    elif file_extension == ".csv":
```
Trymzet, line 378: Should include a comment why you're doing this (why not use `os.stat` and why use `_viadot_downloaded_at_utc` column)

viadot/utils.py
```diff
@@ -345,3 +351,31 @@ def union_dict(*dicts):
 
     """
     return dict(chain.from_iterable(dct.items() for dct in dicts))
+
+
+def handle_if_empty_file(if_empty: Literal["warn", "skip", "fail"] = "warn"):
+    if if_empty == "warn":
+        logger.warning("The input file is empty.")
+    elif if_empty == "skip":
+        raise SKIP()
```
Trymzet, line 360: This could be improved to pass a useful message, eg. by accepting a param `message` which you pass to the warnings/exceptions. Then in `check_if_empty_file()` you can pass the path to the file inside the `message`.

tests/unit/test_utils.py
```diff
@@ -66,3 +71,33 @@ def test_double_quotes_inside():
...
+
+    with caplog.at_level(logging.WARNING):
+        check_if_empty_file(path="empty.csv", if_empty="warn", file_extension=".csv")
+        assert "The input file is empty." in caplog.text
+    with pytest.raises(ValueError):
+        check_if_empty_file(path="empty.csv", if_empty="fail", file_extension=".csv")
+    with pytest.raises(SKIP):
+        check_if_empty_file(path="empty.csv", if_empty="skip", file_extension=".csv")
+
+    os.remove("empty.csv")
```
Trymzet, line 88: You have three different handles inside `check_if_empty_file` I think (CSV, Parquet, and other), so I think this test should handle all three cases (or you can create 3 separate tests if that's easier).

tests/unit/test_utils.py
```diff
@@ -66,3 +71,33 @@ def test_double_quotes_inside():
...
```
Trymzet, line 103: Same as above, we're only testing the CSV here but the Parquet and other file types are handled by a separate code path inside `check_if_empty_file()`, so they should all be tested.

Trymzet (changes requested):

viadot/task_utils.py
```diff
@@ -478,17 +478,15 @@ def concat_dfs(dfs: List[pd.DataFrame]):
 @task
 def map_dtypes_to_str(df: pd.DataFrame) -> pd.DataFrame:
```
Trymzet, line 479: Pls also change the name of the func, eg. `cast_df_to_str()`.

Trymzet (approved):
Looks good :)

PR #441, opened 2022-06-27
✨ Changed  `add_ingestion_metadata_task`  logic

angelika233:
## Summary
-Changed `add_ingestion_metadata_task()` to not to add metadata column when input DataFrame is empty, `check_if_empty_file()` logic according to changes in `add_ingestion_metadata_task()`
- Changed accepted values of `if_empty` parameter in `DuckDBCreateTableFromParquet`

## Importance
There won't be an issue when metadata column is added to empty df

## Checklist

This PR:

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -81,30 +81,12 @@ def test_check_if_empty_file_csv(caplog):
         pass
 
     with caplog.at_level(logging.WARNING):
-        check_if_empty_file(path=EMPTY_CSV_PATH, if_empty="warn", file_extension=".csv")
+        check_if_empty_file(path=EMPTY_CSV_PATH, if_empty="warn")
         assert f"Input file - '{EMPTY_CSV_PATH}' is empty." in caplog.text
     with pytest.raises(ValueError):
-        check_if_empty_file(path=EMPTY_CSV_PATH, if_empty="fail", file_extension=".csv")
+        check_if_empty_file(path=EMPTY_CSV_PATH, if_empty="fail")
     with pytest.raises(SKIP):
-        check_if_empty_file(path=EMPTY_CSV_PATH, if_empty="skip", file_extension=".csv")
-
-    os.remove(EMPTY_CSV_PATH)
-
-
-def test_check_if_empty_file_one_column_csv(caplog):
-    df = pd.DataFrame(columns=["_viadot_downloaded_at_utc"], index=[0])
-    df.to_csv(EMPTY_CSV_PATH, index=False)
-
-    with caplog.at_level(logging.WARNING):
-        check_if_empty_file(path=EMPTY_CSV_PATH, if_empty="warn", file_extension=".csv")
-        assert (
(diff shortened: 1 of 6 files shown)


viadot/tasks/duckdb.py
```diff
@@ -126,7 +126,7 @@ def run(
             NoReturn: Does not return anything.
         """
         try:
-            check_if_empty_file(path=path, if_empty=if_empty, file_extension=".parquet")
+            check_if_empty_file(path=path, if_empty=if_empty)
```
Trymzet, line 129: If you default to `None` for `if_empty` here and you pass it over to `check_if_empty_file()`, it will be `None` in `check_if_empty_file()` (NOT the default value you set in `check_if_empty_file()` -- the default is only used if you don't pass the parameter at all). 

There are two ways to solve this: 
1) Set the default here to the same default as in `check_if_empty_file()`, 
2) Check for `None` in `check_if_empty_file()`, eg. `if_empty = if_empty or "warn"`.

viadot/task_utils.py
```diff
@@ -37,9 +37,14 @@ def add_ingestion_metadata_task(
     Args:
         df (pd.DataFrame): input DataFrame.
     """
-    df2 = df.copy(deep=True)
-    df2["_viadot_downloaded_at_utc"] = datetime.now(timezone.utc).replace(microsecond=0)
-    return df2
+    if len(df.columns) == 0:
```
Trymzet, line 40: Or just `if not df.columns` which is closer to natural language ("if there are no columns"). Also, please add a comment explaining why we do this check as it's not self-evident (ie. why we can't use df.empty here).

tests/unit/test_utils.py
```diff
@@ -114,41 +96,11 @@ def test_check_if_empty_file_parquet(caplog):
...
-        )
-    with pytest.raises(SKIP):
-        check_if_empty_file(
-            path=EMPTY_PARQUET_PATH, if_empty="skip", file_extension=".parquet"
-        )
-
-    os.remove(EMPTY_PARQUET_PATH)
-
-
-def test_check_if_empty_file_one_column_parquet(caplog):
```
Trymzet, line 133: Can you pls add a test for the case with a parquet file with columns but no data?

Trymzet (changes requested):

viadot/tasks/duckdb.py
```diff
@@ -105,7 +105,7 @@ def run(
         path: str,
         schema: str = None,
         if_exists: Literal["fail", "replace", "append", "skip", "delete"] = None,
-        if_empty: Literal["warn", "skip", "fail"] = None,
```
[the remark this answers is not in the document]
angelika233: When we pass empty file to this task we get an error, so warning doesn't make sense

viadot/task_utils.py
angelika233: `if not df.columns` causes an error so I went back to `if len(df.columns) == 0`

Trymzet: Interesting, what's the error?

Trymzet (approved):
LGTM

PR #452, opened 2022-07-12
Added rollback feature, improved comments formatting

afraijat:
## Summary

Added the rollback feature to the Databricks source.

## Importance

The ability to rollback is crucial during testing, and is a great feature in general, in case an incorrect/accidental operation is done on the table.

## Checklist

This PR:

### DIFF
--- a/tests/integration/tasks/test_databricks.py
+++ b/tests/integration/tasks/test_databricks.py
@@ -112,7 +112,7 @@ def test_insert_wrong_schema():
 
     with pytest.raises(pyspark.sql.utils.AnalysisException):
         did_insert = databricks.insert_into(
-            SCHEMA, table=table, df=append_df, if_exists="append"
+            schema=SCHEMA, table=table, df=append_df, if_exists="append"
         )
 
 
--- a/tests/integration/tasks/test_databricks.py
+++ b/tests/integration/tasks/test_databricks.py
@@ -4,7 +4,7 @@ import pandas as pd
 import pytest
 
 databricks = Databricks(env="QA")
-schema = "afraijat_test"
+SCHEMA = "afraijat_test"
 
 source_data = [
     {
@@ -32,7 +32,7 @@ df = pd.DataFrame(source_data)
 def test_to_df():
     # Look up test_query table and check if the type and results returned are correct
     table = "test_query"
-    fqn = f"{schema}.{table}"
+    fqn = f"{SCHEMA}.{table}"
 
     result = databricks.to_df(f"select * from {fqn}")
 
@@ -43,18 +43,18 @@ def test_create_table():
 
     table = "test_create_table"
 
-    fqn = f"{schema}.{table}"
+    fqn = f"{SCHEMA}.{table}"
 
-    exists = databricks._check_if_table_exists(schema, table)
+    exists = databricks._check_if_table_exists(SCHEMA, table)
 
     assert not exists
 
-    databricks.create_table_from_pandas(schema, table, df)
+    databricks.create_table_from_pandas(SCHEMA, table, df)
 
     result = databricks.to_df(f"SELECT * FROM {fqn}")
 
     assert result.shape == df.shape
-    databricks.drop_table(schema, table)
+    databricks.drop_table(SCHEMA, table)
 
 
 def test_append():
@@ -72,14 +72,14 @@ def test_append():
 
     table = "test_append_table"
 
-    fqn = f"{schema}.{table}"
+    fqn = f"{SCHEMA}.{table}"
 
-    databricks.create_table_from_pandas(schema, table, df)
+    databricks.create_table_from_pandas(SCHEMA, table, df)
 
     append_df = pd.DataFrame(append_data)
 
     did_insert = databricks.insert_into(
-        schema, table=table, df=append_df, if_exists="append"
+        SCHEMA, table=table, df=append_df, if_exists="append"
     )
     assert did_insert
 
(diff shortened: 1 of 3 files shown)


viadot/sources/databricks.py
```diff
@@ -428,3 +428,61 @@ def discover_schema(self, schema: str, table: str) -> dict:
...
+        return ver_num
+
+    def rollback(self, schema: str, table: str, ver_num: int):
+        """
+        Rollback the table to a previous version.
+
+        Args:
+            schema (str): Name of the schema.
+            table (str): Name of the table to rollback.
+            ver_num (int): Number of the table's version to rollback to.
```
Trymzet, line 460: use the full name for clarity: `version_number`

viadot/sources/databricks.py
```diff
@@ -428,3 +428,61 @@ def discover_schema(self, schema: str, table: str) -> dict:
...
+
+        Example:
+        ```python
+        from viadot.sources import Databricks
+        databricks = Databricks()
+
+        schema = "test"
+        table = "table_1"
+
+        ver_num = databricks.get_table_ver(schema = schema, table = table)
```
Trymzet, line 470: In Python, there is no space around keyword arguments, so it's `func(a=a)` instead of `func(a = a)`.

viadot/sources/databricks.py
```diff
@@ -428,3 +428,61 @@ def discover_schema(self, schema: str, table: str) -> dict:
...
+
+        # Perform changes on the table, in this example we are appending to the table
+        list = [{"id":"1", "name":"Joe"}]
+        df = pd.DataFrame(list)
+        databricks.insert_into(schema = schema, table = table, df = df, if_exists="append")
```
Trymzet, line 475: same as above

viadot/sources/databricks.py
```diff
@@ -428,3 +428,61 @@ def discover_schema(self, schema: str, table: str) -> dict:
...
+
+        databricks.rollback(schema, table, ver_num)
+        ```
+        """
+        fqn = f"{schema}.{table}"
```
Trymzet, line 480: This will break if schema is None. You should check for that and use `Databricks.DEFAULT_SCHEMA` in such case.

tests/integration/tasks/test_databricks.py
```diff
@@ -230,3 +230,41 @@ def test_discover_schema():
...
+            "Name": "new upsert-2",
+            "FirstName": "Updated",
+            "LastName": "Carter2",
+            "ContactEMail": "Adam.Carter@TurnerBlack.com",
+            "MailingCity": "Updated!Jamesport",
+        }
+    ]
+
+    table = "test_rollback"
+    fqn = f"{schema}.{table}"
```
Trymzet, line 249: There is a convention to use CAPS for constants in Python. In this case, `SCHEMA` is a constant as you're using the same schema for all tests.

Trymzet (changes requested):

viadot/sources/databricks.py
```diff
@@ -413,7 +415,7 @@ def discover_schema(self, schema: str, table: str) -> dict:
         databricks.discover_schema(schema = "schema", table = "table")
         ```
         Returns:
-            schema: A dictionary containing the schema details of the table.
+            schema (dict): A dictionary containing the schema details of the table.
         """
         fqn = f"{schema}.{table}"
         result = self.run(f"DESCRIBE {fqn}", fetch_type="pandas")
```
Trymzet, line 421: Looks like this one is also missing a guard against `schema` being None.

viadot/sources/databricks.py
```diff
@@ -428,3 +430,74 @@ def discover_schema(self, schema: str, table: str) -> dict:
         schema = dict(zip(col_names, data_types))
 
         return schema
+
+    def get_table_version(self, table: str, schema: str = None) -> numpy.int64:
```
Trymzet, line 434: Is this really a numpy type and not a standard Python int? If so, can we make it a standard Python int? I think if we can have a simpler type, we should, and version number is a simple small number than can easily be stored as int. 

You probably have to convert it with int(your_numpy_number) but for me it's worth it when the brain doesn't have to stop and think "wait, why is a version number stored as a a numpy thingy?" 😛.

viadot/sources/databricks.py
```diff
@@ -428,3 +430,74 @@ def discover_schema(self, schema: str, table: str) -> dict:
...
+        """
+        Get the provided table's version number.
+
+        Args:
+            schema (str): Name of the schema.
+            table (str): Name of the table to rollback.
+
+        Returns:
+            version_number (numpy.int64): The table's version number
```
Trymzet, line 443: same as above

Trymzet (changes requested):

viadot/sources/databricks.py
```diff
@@ -199,18 +201,21 @@ def run(
 
         return result
 
-    def _check_if_table_exists(self, schema: str, table: str) -> bool:
+    def _check_if_table_exists(self, table: str, schema: str = None) -> bool:
+        if schema is None:
+            schema = Databricks.DEFAULT_SCHEMA
```
Trymzet, line 206: This depends on `Databricks.DEFAULT_SCHEMA` which is not defined

Trymzet (changes requested):

viadot/sources/databricks.py
afraijat: Kindly refer to line 43

Trymzet: Ah nice, I missed it. While we're at it though you can remove the other two variables from there (`env` is already provided in `__init__` and `session` should be initialized in `__init__` as well):
```python
        self.env = env
        self.credentials = credentials or local_config.get("DATABRICKS", {}).get(env)
        self.session = None

        self.connect()
```

Trymzet (approved):
Looks good :)

PR #460, opened 2022-07-20
🐛 Fixed bug while appending to non existing table

angelika233:
## Summary
Changed `ADLSToAzureSQL` flow not to map if_exist while checking column order. Changed if structure in `CheckColumnOrder`.

## Importance
Fixed a bug

## Checklist

This PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `on_bcp_error` and `bcp_error_log_path` parameters in `BCPTask`
 - Added ability to process queries which result exceed SAP's character per low limit in `SAPRFC` source
 ### Changed
+- Cahnged `CheckColumnOrder` and `ADLSToAzureSQL` to handle appending to non existing table
 - Changed `add_ingestion_metadata_task()` to not to add metadata column when input DataFrame is empty
 - Changed `check_if_empty_file()` logic according to changes in `add_ingestion_metadata_task()`
 - Changed accepted values of `if_empty` parameter in `DuckDBCreateTableFromParquet`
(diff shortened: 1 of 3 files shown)


viadot/tasks/azure_sql.py
```diff
@@ -407,10 +407,10 @@ def run(
         query = f"SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = '{schema}' AND TABLE_NAME = '{table}'"
         check_result = azure_sql.run(query=query)
         if if_exists not in ["replace", "fail"]:
-            if if_exists == "append" and not check_result:
+            if if_exists == "append" and len(check_result) == 0:
                 self.logger.warning("Aimed table doesn't exists.")
```
Trymzet, line 411: 1. Should be: "target table"
2. This whole block from line 408 could be refactored as it's quite confusing:
- ` if if_exists not in ["replace", "fail"]:` - it's easier to read if you do `if if_exists in ["append", "delete"]`
- you should add a check that `if_exists` is one of ["fail", "replace", "append", "delete"] before this block. Then it's also clear you don't need to check for that possibility here.
- `len(check_result) == 0` - you could set this as a variable, eg. `table_exists =  len(check_result) == 0`. This way this line is more intuitive: `if if_exists == "append" and not table_exists: (fail)`. It seems like the lines below could also be simplified if you utilize this `table_exists` variable, because eg. on line 415 you iterate over the results even though you're not sure the table even exists (you only check for "doesn't exist" + "append" combination on line 410 but not for "doesn't exist" + "delete").
- `result = azure_sql.run(query=query)` (line 414) -- seems to duplicate the call made on line 408
- `self.logger.info("The table will be replaced.")` (line 426) -- shouldn't it fail right when you discover that the table exists and if_exists is set to "fail"? So `if table_exists and if_exists == "fail": raise sth()`?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `on_bcp_error` and `bcp_error_log_path` parameters in `BCPTask`
 - Added ability to process queries which result exceed SAP's character per low limit in `SAPRFC` source
 ### Changed
+- Cahnged `CheckColumnOrder` and `ADLSToAzureSQL` to handle appending to non existing table
```
Trymzet, line 12: - typo in "Changed"
- I would add "task" and "flow" words to be a bit more explicit, so "Changed `CheckColumnOrder` task and `ADLSToAzureSQL` flow ..."

Trymzet (changes requested):
Some refactoring of the `run()` method required

PR #493, opened 2022-08-16
Add `find` path support to AzureDataLake class

winiar93:
## Summary
Now `AzureDataLake` class contains `find()` method that list all files and folders inside provided patch.
Also we modified `AzureDataLakeList` task and now it's able to show recursive directory.

## Importance
It may be important because while data ignestion process e.g if there will be situation like:
`c4c/account/2022/06/27/account.csv`
`c4c/account/2022/06/27/tickets.csv`
`c4c/account/2022/06/28/account.csv`
`c4c/account/2022/06/28/tickets.csv`
`c4c/account/2022/06/29/account.csv`
`c4c/account/2022/06/30/account.csv`

and we would like to load only account.csv file we can genereate list of path for file with specific name.

`list_in_adls = AzureDataLakeList()
valid_paths = list_in_adls.run(
    path="raw/tests/alds_test_new_fnc",
    recursive=True,
    file_to_match = 'final_df2.csv',
    sp_credentials_secret="*****",
)`

OUTPUT:
['raw/tests/alds_test_new_fnc/2020/02/01/final_df2.csv',
 'raw/tests/alds_test_new_fnc/2020/02/02/final_df2.csv',
 'raw/tests/alds_test_new_fnc/2021/12/01/final_df2.csv',
 'raw/tests/alds_test_new_fnc/2022/06/21/final_df2.csv',
 'raw/tests/alds_test_new_fnc/2022/08/12/final_df2.csv',
 'raw/tests/alds_test_new_fnc/test_folder/final_df2.csv',
 'raw/tests/alds_test_new_fnc/test_folder_2/final_df2.csv']

after that we can iterate over list and run `ADLSToAzureSQL` flow on each iteration.

## Checklist

This PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added handling empty DF in `set_new_kv()` task
 - Added `update_kv` and `filter_column` params to `SAPRFCToADLS` and `SAPToDuckDB` flows and added `set_new_kv()` task
+- Added `recursive` parameter to `AzureDataLakeList` task
 in `task_utils`
 ## [0.4.6] - 2022-07-21
 ### Added
--- a/tests/integration/tasks/test_azure_data_lake.py
+++ b/tests/integration/tasks/test_azure_data_lake.py
@@ -24,6 +24,23 @@ ADLS_PATH_2 = f"raw/supermetrics/{FILE_NAME_2}"
 FILE_NAME_PARQUET = f"test_file_{uuid_4}.parquet"
 ADLS_PATH_PARQUET = f"raw/supermetrics/{FILE_NAME_PARQUET}"
 
+ADLS_TEST_PATHS = [
+    "raw/tests/alds_test_new_fnc/2020/02/01/final_df2.csv",
+    "raw/tests/alds_test_new_fnc/2020/02/01/test_new_fnc.csv",
+    "raw/tests/alds_test_new_fnc/2020/02/02/final_df2.csv",
+    "raw/tests/alds_test_new_fnc/2020/02/02/test_new_fnc.csv",
+    "raw/tests/alds_test_new_fnc/2021/12/01/final_df2.csv",
+    "raw/tests/alds_test_new_fnc/2021/12/01/test_new_fnc.csv",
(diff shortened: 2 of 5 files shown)


Trymzet:
Sorry, the title and description are confusing. The title mentions upload, but in the description you talk about listing files. Can you fix one or the other before we jump into code review?

winiar93:
> Sorry, I find the title and description confusing. The title mentions upload, but in the description you talk about listing files. Can you fix one or the other before we jump into code review?

Description updated.

Trymzet:
I think you should explain in the description why you're not using the intuitive solution of `adlfs.glob()` or any other built-in functions from `adlfs` or the Python `glob` lib (in case previous ones don't work somehow). 

To me it seems like the feature you're trying to implement is "Add `glob`-style path support to `AzureDataLake.ls()`", so eg. the ability for the user to specify the path as `adls_path="my/path/*"`. I don't think this should require a separate parameter.

Diego-H-S:
> I think you should explain in the description why you're not using the builtin `ls()` or things like `adlfs.glob()` or any other built-in functions from `adlfs` or the Python `glob` lib (in case previous ones don't work somehow).

that's right 
 
> To me it seems like the feature you're trying to implement is "Add `glob`-style path support to `AzureDataLake.ls()`", so eg. the ability for the user to specify the path as `adls_path="my/path/*"`. I don't think this should require a separate parameter in `AzureDataLakeList`. It's even possible that `ls` already handles recursiveness by default, so eg, if you do `adls.ls("path/")` it will already list all the files in that path.

I have read the source code of Azure, and the `ls()` method doesn't implement recursiveness (or at least I didn't see that feature), and I didn't find any other built-in function that returns a list of paths/files deeper than the one that you pass as an argument.

Trymzet:
> I have read the source code of Azure, and the `ls()` method doesn't implement recursiveness (or at least I didn't see that feature), and I didn't find any other built-in function that returns a list of paths/files deeper than the one that you pass as an argument.

What about `glob()`?

Also, could you please add the explanation to the PR description?

Diego-H-S:
> What about `glob()`?

Not `glob()` but `find()`

Rafalz13 (commented):
Looks good

Rafalz13 (approved):
@m-paz

PR #518, opened 2022-09-20
🐛 Fixed  `if_exists="delete"` in `DuckDB()`

angelika233:
## Summary
Fixed incorrect `if_exists="delete"` handling in `DuckDB.create_table_from_parquet()`

## Importance
Fixed a bug

## Checklist

This PR:


viadot/sources/duckdb.py
```diff
@@ -172,6 +172,9 @@ def create_table_from_parquet(
                 return True
             elif if_exists == "delete":
                 self.run(f"DELETE FROM {fqn}")
+                self.logger.info(f"Successfully deleted data from table '{fqn}'.")
+                self.run(f"INSERT INTO {fqn} SELECT * FROM read_parquet('{path}')")
+                self.logger.info(f"Successfully inserted data into table '{fqn}'.")
```
Trymzet, line 177: Does this make sense here considering we run below just after these lines? Don't these lines override your code? It seems if we always run a CTAS, then the "delete" option doesn't make sense at all since we're always creating a new table from scratch.
https://github.com/dyvenia/viadot/blob/d73d6e5ff698b4dfae54c7cde2fb1d65635cf3d7/viadot/sources/duckdb.py#L189-L190

tests/unit/test_duckdb.py
```diff
@@ -56,6 +56,25 @@ def test_create_table_from_parquet_append(duckdb, TEST_PARQUET_FILE_PATH):
...
+    duckdb.create_table_from_parquet(
+        schema=SCHEMA, table=TABLE, path=TEST_PARQUET_FILE_PATH, if_exists="delete"
+    )
+    df = duckdb.to_df(f"SELECT * FROM {SCHEMA}.{TABLE}")
+    assert df.shape[0] == 3
+
+    duckdb.drop_table(TABLE, schema=SCHEMA)
+    duckdb.run(f"DROP SCHEMA {SCHEMA}")
+
+
```
Trymzet, line 77: One way to test delete specifically is to load some data with a new schema and make sure this load fails.

Trymzet (changes requested):
Some questions

viadot/sources/duckdb.py
angelika233: In this case we are not creating table because we finish code execution with return in line 178

tests/unit/test_duckdb.py
Trymzet: Once you resolve this one, pls resolve the conflicts as well

Trymzet (approved):

PR #548, opened 2022-10-18
✨ Added parameters to functions in config.py

djagoda881:
## Summary

Adding an additional parameter `config` to functions `get_source_config` and `get_source_credentials`

## Importance

Allows you to add your own `config` to the function

## Checklist

This PR:

### DIFF
--- a/viadot/config.py
+++ b/viadot/config.py
@@ -36,15 +36,15 @@ except FileNotFoundError:
         CONFIG = Config()
 
 
-def get_source_config(key):
-    source_configs = CONFIG.get("sources")
+def get_source_config(key, config):
+    source_configs = config.get("sources")
     if source_configs is not None:
         for source_config in source_configs:
             if key in source_config.keys():
                 return source_configs[source_configs.index(source_config)][key]
 
 
-def get_source_credentials(key):
-    config = get_source_config(key)
+def get_source_credentials(key, config):
+    config = get_source_config(key, config)
     if config is not None:
         return config.get("credentials")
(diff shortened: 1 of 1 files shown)


viadot/config.py
```diff
@@ -36,15 +36,15 @@ def from_yaml(cls, path: str, key: Optional[str] = None) -> Config:
         CONFIG = Config()
 
 
-def get_source_config(key):
-    source_configs = CONFIG.get("sources")
+def get_source_config(key, config):
```
Trymzet, line 39: `config` should be optional as otherwise you would break all existing sources that use this function

viadot/config.py
```diff
@@ -36,15 +36,15 @@ def from_yaml(cls, path: str, key: Optional[str] = None) -> Config:
...
+    source_configs = config.get("sources")
     if source_configs is not None:
         for source_config in source_configs:
             if key in source_config.keys():
                 return source_configs[source_configs.index(source_config)][key]
 
 
-def get_source_credentials(key):
-    config = get_source_config(key)
+def get_source_credentials(key, config):
```
Trymzet, line 47: `config` should be optional as otherwise you would break all existing sources that use this function

viadot/config.py
Trymzet: you can set `CONFIG` as default

Trymzet (changes requested):

viadot/config.py
djagoda881: Done

viadot/config.py
djagoda881: Done

PR #557, opened 2022-10-28
Sap rfc source

Diego-H-S:
## Summary
Bug fixed in `sap_rfc.py` source, which was failing by new rows in SAP table between iterations, and also fixed the flow breakdown when a separator character was inside the string text retrieved from the response.

## Importance
Very important, because in such SAP tables where the updating time is very low, the flow was failing almost every run. Also important, because when the flow was broken by an extra separator character, the whole day (and the next one) is lost.

## Checklist

This PR:


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added timeout to `DuckDBQuery` and `SAPRFCToDF`
 - Added support for SQL queries with comments to `DuckDB` source
 - Added "WITH" to query keywords in `DuckDB` source
+- Added new parameter to `sap_rfc_to_adls.py` to replace a separator character 
+inside a string variable by defined character.
```
Trymzet, line 13: I find it a bit hard to follow precisely, maye you can name the parameter here? So eg. "Added new paramter `x` to `y` ..."

CHANGELOG.md
```diff
@@ -17,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed incorrect `if_exists="delete"` handling in `DuckDB.create_table_from_parquet()`
 - Fixed `test_duckdb_to_sql_server.py` tests - revert to a previous version
 - Removed `test__check_if_schema_exists()` test
+- Fixed `sap_rfc.py` source file to not breakdown by both, 
+and extra separator in a row and adding new rows in SAP table between iterations.
```
Trymzet, line 23: Sorry I don't understand:
- what does "not breakdown by both" mean here?
- the second part I'm having trouble following at all, something is happening between iterations of something?

viadot/flows/sap_rfc_to_adls.py
```diff
@@ -50,6 +51,8 @@ def __init__(
             name (str): The name of the flow.
             query (str): Query to be executed with pyRFC. Defaults to None.
             rfc_sep(str, optional): Which separator to use when querying SAP. If not provided, multiple options are automatically tried.
+            rfc_replacement (str, optional): In case of sep is on a columns, set up a new character to replace
+                inside the string to avoid flow breakdowns. Defaults to "-".
```
Trymzet, line 55: Ah, so you mean in case the character provided in `rfc_sep` is present in a column, you replace that char with `rfc_replacement`? Normally, the way to solve this problem is by using a concept called "escaping". For example, this is how you escape special characters in Python: https://www.w3schools.com/python/gloss_python_escape_characters.asp. Have you tried doing it that way?

viadot/sources/sap_rfc.py
```diff
@@ -463,37 +468,86 @@ def to_df(self):
...
+                    response = self.call(func, **params)
+                except ABAPApplicationError as e:
+                    if e.key == "DATA_BUFFER_EXCEEDED":
+                        raise DataBufferExceeded(
+                            "Character limit per row exceeded. Please select fewer columns."
+                        )
+                    else:
+                        raise e
+                record_key = "WA"
+                data_raw = np.array(response["DATA"])
```
Trymzet, line 490: Let's avoid using numpy where standard Python is enough.

viadot/sources/sap_rfc.py
```diff
@@ -463,37 +468,86 @@ def to_df(self):
...
+                        logger.warning(
+                            f"Empty output was generated for chunk {chunk} in columns {fields}."
+                        )
+                        chunk += 1
+                        continue
+                elif data_raw.shape[0] != row_index:
+                    data_raw = data_raw[:row_index]
+                    logger.warning(
+                        f"New rows were generated during the execution of the script. The table is truncated to the number of rows for the first chunk"
+                    )
```
Trymzet, line 504: What is this new logic for? There should be a comment at the top explaining it since it's not self-evident.

viadot/sources/sap_rfc.py
```diff
@@ -463,37 +468,86 @@ def to_df(self):
...
+
+                # first we identify where the data has an extra sep due to text
+                sep_counts = np.array([], dtype=int)
```
Trymzet, line 507: I don't think `numpy` is needed here or below either, seems like you can do that with regular Python lists.

But anyway take a step back and think if this problem can be solved using escape characters since you're adding a lot of custom logic here just to basically do that. 

Also, please split the logic into functions so that it's easier to follow (and also add docstrings to those functions explaining what each of them is doing + typing). Looking at the code below it should probably be ~5 functions. The for loop as it is now has ~80 lines of code; it shouldn't be more than 10-20.

Once you rewrite (or use escape character approach) I'll be able to provide some feedback on the logic.

Trymzet (changes requested):
I'm not sure the approach is correct, but even if it is, the implementation requires refactoring the logic into functions. Once that is done I can also provide more feedback since now I'm not able to follow the logic without significant effort.

CHANGELOG.md
Diego-H-S: I have improved the file, check the structure and tell me if it is better now.

Diego-H-S (commented):
I have implemented all the changes except for the numpy uses, that I consider it faster, easier to work, and compact.

CHANGELOG.md
Diego-H-S: Changed the structure.

viadot/sources/sap_rfc.py
Diego-H-S: done, have a look at it.

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+    # indentifying "good" rows we obtain the index of separatos positions.
+    pos_sep_index = np.array([], dtype=int)
+    for data in data_raw[sep_index]:
+        pos_sep_index = np.append(
+            pos_sep_index,
+            np.where(np.array([*data[record_key]]) == f"{sep}"),
+        )
+    pos_sep_index = np.unique(pos_sep_index)
+
+    # in rows with an extra separator, we replace them by another character: "-" by default
```
Rafalz13, line 166: The function name is `catch_extra_separator`. Maybe it is a good idea to create new function because this one seems to be complex. You can extract the `replace_separator_in_data` code block from this function and create a new one.

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+        row_index (int): Number of rows set it down in he first iteration with the SAP table.
+        data_raw (np.array): Array with the data retrieve from SAP table.
+        chunk (int): The part number in which a number of SAP table columns have been split.
+        fields (List[str]): A list with the names of the columns in a chunk.
+
+    Returns:
+        Union[int, np.array, bool]: A tuple with the parameters "row_index", "data_raw", and a new
+            boolean variable "cont" to indicate when the for loop has to be restarted.
+    """
+    cont = False
```
Rafalz13, line 104: The `cont` name is not clear to me. Can you add a different name for this variable if possible? To make the code easier to read.

Rafalz13 (changes requested):
Added comments and suggestions.

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+
+
+def catch_extra_separators(
+    data_raw: np.array, record_key: str, sep: str, fields: List[str], replacement: str
+) -> np.array:
+    """Function to replace extra separators in every row of the
+
+    Args:
+        data_raw (np.array): Array with the data retrieve from SAP table.
+        record_key (str): Key word to extract the data from the numpy array "data_raw".
```
angelika233, line 129: This sentence is not clear to me maybe paraphrase it

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+    Returns:
+        np.array: The argument "data_raw" with no extra delimiters.
+    """
+
+    # remove scape characters from data_raw ("\t")
+    for n, r in enumerate(data_raw):
+        if "\t" in r[record_key]:
+            data_raw[n][record_key] = r[record_key].replace("\t", " ")
+
+    # first we identify where the data has an extra separator in text columns.
```
angelika233, line 144: maybe change comments from "we" to something more generic

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+    sep_counts = np.array([], dtype=int)
+    for row in data_raw:
+        sep_counts = np.append(sep_counts, row[record_key].count(f"{sep}"))
```
angelika233, line 147: In different files we mostly use pandas,would you be able to replace numpy with pandas?

viadot/sources/sap_rfc.py
```diff
@@ -438,7 +580,6 @@ def to_df(self):
             pd.DataFrame: A DataFrame representing the result of the query provided in `PyRFC.query()`.
         """
         params = self._query
-        columns = self.select_columns_aliased
```
angelika233, line 441: you cannot remove this part (and code below) because without it columns dont get aliased

angelika233 (changes requested):
The most important change:
-After your changes columns don't get aliased [VERY IMPORTANT]
-Those changes should be optional, older option should be a default
Suggestion, but would be good if can be applied:
-Change numpy to pandas 
-Maybe some time optimization, it runs few times longer than before (I will also think about it and will let you know if I have any idea :) )

CHANGELOG.md
```diff
@@ -5,8 +5,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 ### Added
 - Added `anonymize_df` task function to `task_utils.py` to anonymize data in the dataframe in selected columns.
+- Added new parameter `rfc_replacement` to `sap_rfc_to_adls.py` to replace 
+an extra separator character within a string column to avoid conflicts.
+- Added `rfc_reference_column` in `sap_frc.py` to merge chunks on this column.
+
+### Fixed
+- Removed `try-except` sentence and added a new logic to remove extra separators in `sap_rfc.py` 
+source file, to vaoid a mismatch in columns lenght between iterative connections to SAP tables.
+- When `SAP` tables are updated during `sap_rfc.py` scrip running, if there are chunks, the
```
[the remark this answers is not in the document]
Diego-H-S: removed lines 18 and 19 because it just expand what is said in the previous lines.

CHANGELOG.md
```diff
@@ -66,6 +76,8 @@ This parameter enables user to decide whether or not filter should be validated.
 - Fixed incorrect `if_exists="delete"` handling in `DuckDB.create_table_from_parquet()`
 - Fixed `test_duckdb_to_sql_server.py` tests - revert to a previous version
 - Removed `test__check_if_schema_exists()` test
+- Fixed `sap_rfc.py` source file to not breakdown by both,
```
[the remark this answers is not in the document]
Diego-H-S: some remanents from git rebase :-D. done!

viadot/flows/sap_rfc_to_adls.py
```diff
@@ -48,8 +49,11 @@ def __init__(
...
             func (str, optional): SAP RFC function to use. Defaults to "RFC_READ_TABLE".
             rfc_total_col_width_character_limit (int, optional): Number of characters by which query will be split in chunks in case of too many columns
+            rfc_reference_column (List[str], optional): Reference columns to merge chunks Data Frames. These columns must to be unique. Defaults to None.
```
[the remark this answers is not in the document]
Diego-H-S: name changed and add more docstrings.

viadot/tasks/sap_rfc.py
```diff
@@ -15,6 +16,7 @@ def __init__(
         self,
         query: str = None,
         sep: str = None,
+        replacement: str = "-",
```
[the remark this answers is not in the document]
Diego-H-S: Not necessary to do that in `init()` because it is already explained in `run()`. However, if you think should be recommended do it in `init()`, I could do it.

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+            inside the string to avoid flow breakdowns.
+
+    Returns:
+        np.array: The argument "data_raw" with no extra delimiters.
+    """
+
+    # remove scape characters from data_raw ("\t")
+    for n, r in enumerate(data_raw):
+        if "\t" in r[record_key]:
+            data_raw[n][record_key] = r[record_key].replace("\t", " ")
```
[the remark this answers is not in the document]
Diego-H-S: must be a string of length 1 to avoid changing the total length of the string. I think the space character is the less information provider.

viadot/sources/sap_rfc.py
```diff
@@ -113,11 +213,14 @@ def __init__(
         Args:
             sep (str, optional): Which separator to use when querying SAP. If not provided,
             multiple options are automatically tried.
+            replacement (str, optional): In case of sep is on a columns, set up a new character to replace
```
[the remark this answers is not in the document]
Diego-H-S: done.

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+
+    no_sep_index = np.argwhere(sep_counts != len(fields) - 1)
+    no_sep_index = no_sep_index.reshape(
+        len(no_sep_index),
+    )
+    sep_index = np.argwhere(sep_counts == len(fields) - 1)
+    sep_index = sep_index.reshape(
+        len(sep_index),
+    )
+    # indentifying "good" rows we obtain the index of separatos positions.
```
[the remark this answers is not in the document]
Diego-H-S: yes it is.

viadot/sources/sap_rfc.py
Diego-H-S: I changed it to `start`, I think it should be fine, but I accept suggestions :)

viadot/sources/sap_rfc.py
Diego-H-S: I think all the code in that function should stay together because everything is related. Maybe I don't understand exactly what you want to say

viadot/sources/sap_rfc.py
Diego-H-S: done.

viadot/sources/sap_rfc.py
Diego-H-S: Not all the functionalities implemented with `numpy` can be done with `pandas`. For now, I don't think that is necessary, furthermore, it is faster.

viadot/sources/sap_rfc.py
Diego-H-S: Yes, it can be, because there is no `try-except` sentence anymore, and then if nothing is in `report` the `df` is just populated with nothing but it isn't causing any error. Therefore, not needed lines from 493 to 496 (old code), and then the `columns` variable is useless.

viadot/sources/sap_rfc.py
Diego-H-S: Sorry for that, could you suggest a sentence, I'm not very good at explaining such kinds of things.

viadot/sources/sap_rfc.py
angelika233: I mean this code doesn't give an error but it doesn't work as it's supposed to i.e. when you want to alias column in query "Select abcd as column1" the name of column 'abcd' in df doesnt change to 'column1'

Diego-H-S: Now I understand what you want to point out with that. Changes done!

Diego-H-S (commented):
Please, have a look at all the changes that I implemented by your feedback.
Thanks for improving it!

viadot/sources/sap_rfc.py
Rafalz13: I meant to separate this part of the code and instead of a block of code call a new function here. This part of the functionality would be named for example `replace_separator_in_data` (or in a similar way). Then `catch_extra_separators` will be shorter and (I hope) easier to read. Creating smaller functions can help us, for example, catch errors faster in the future.
What do you think?

Diego-H-S: done!

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,137 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+
+def replace_separator_in_data(
+    data_raw: np.array,
+    no_sep_index: np.array,
+    record_key: str,
+    pos_sep_index: np.array,
+    sep: str,
+    replacement: str,
+) -> np.array:
+    """Fundtion to replace the extra separator in every row of the data_raw numpy array.
```
Rafalz13, line 130: Typo -> `Function`

viadot/sources/sap_rfc.py
```diff
@@ -132,9 +269,12 @@ def __init__(
         super().__init__(*args, credentials=credentials, **kwargs)
 
         self.sep = sep
+        self.replacement = replacement
         self.client_side_filters = None
         self.func = func
         self.rfc_total_col_width_character_limit = rfc_total_col_width_character_limit
+        # remove repeated reference columns
+        self.rfc_unique_id = list(np.unique(np.array(rfc_unique_id)))
```
Rafalz13, line 277: Is it required to use `numpy` here? There is an easy way to get unique values e.g. `list(set(rfc_unique_id))`

viadot/sources/sap_rfc.py
Rafalz13, line 1: You have this 'checks' in `if` -> `isinstance(self.rfc_unique_id[0], str)` in 6 places. Please double-check if this is needed in every place. Maybe this is a way to simplify it and reduce code.

Rafalz13 (changes requested):
- Please check comments from previous review.
- As we discussed - please add new source. The name can be like`SAPRFC2` or `SAPRFCExtended`.
- Add arguments in flows and tasks that allow you to choose what version of source will be used- as default `='SAPRFC'`

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,137 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+        logger.warning("\n" + data_raw[no_sep][record_key])
+        split_array = np.array([*data_raw[no_sep][record_key]])
+        position = np.where(split_array == f"{sep}")[0]
+        index_sep_index = np.argwhere(np.in1d(position, pos_sep_index) == False)
+        index_sep_index = index_sep_index.reshape(
+            len(index_sep_index),
+        )
+        split_array[position[index_sep_index]] = replacement
+        data_raw[no_sep][record_key] = "".join(split_array)
+        logger.warning("\n" + data_raw[no_sep][record_key])
```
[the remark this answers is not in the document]
Diego-H-S: That is for clarity when it would be printed on the screen, of course, it is not necessary but I would recommend that just to better read it on the screen., when the error raises.

viadot/sources/sap_rfc.py
Diego-H-S: done!

viadot/sources/sap_rfc.py
Diego-H-S: implemented!

viadot/sources/sap_rfc.py
Diego-H-S: all of them are necessary because all of them are due to the optional merging column, and it is applied in different parts of the code...

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
     return where_trimmed, wheres_to_add
 
 
+def detect_extra_rows(
+    row_index: int, data_raw: np.array, chunk: int, fields: List[str]
+) -> Union[int, np.array, bool]:
+    """check if, in between calls to the SAP table, the number of rows have increased.
```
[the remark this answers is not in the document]
Diego-H-S: done!

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+        if row_index == 0:
+            logger.warning(
+                f"Empty output was generated for chunk {chunk} in columns {fields}."
+            )
+            chunk += 1
+            cont = True
+    elif data_raw.shape[0] != row_index:
+        data_raw = data_raw[:row_index]
+        logger.warning(
+            f"New rows were generated during the execution of the script. The table is truncated to the number of rows for the first chunk"
```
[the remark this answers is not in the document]
Diego-H-S: added the point ;-)

viadot/sources/sap_rfc.py
```diff
@@ -84,6 +85,103 @@ def trim_where(where: str) -> Tuple[str, OrderedDictType[str, str]]:
...
+        )
+
+    return row_index, data_raw, cont
+
+
+def catch_extra_separators(
+    data_raw: np.array, record_key: str, sep: str, fields: List[str], replacement: str
+) -> np.array:
+    """Function to replace extra separators in every row of the
```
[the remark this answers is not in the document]
Diego-H-S: updated

viadot/tasks/sap_rfc.py
```diff
@@ -68,41 +73,49 @@ def __init__(
...
                 in case of too many columns for RFC function. According to SAP documentation, the limit is
                 512 characters. However, we observed SAP raising an exception even on a slightly lower number
                 of characters, so we add a safety margin. Defaults to None.
+            rfc_reference_column (List[str], optional): Reference columns to merge chunks Data Frames. These columns must to be unique. Defaults to None.
         """
         if query is None:
             raise ValueError("Please provide the query.")
+        if rfc_reference_column:
+            self.logger.warning(
+                "Reference column added. Remember, these columns MUST BE UNIQUE or the table will be malformed."
```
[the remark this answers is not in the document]
Diego-H-S: changed

Diego-H-S (commented):
I have created a second source class to avoid its implementation by default.

Rafalz13 (commented):
- Please rebase because there is new code (for SAPRFC) on DEV branch.

- Please change the name from `new_approx` to something clearer (`new_approximation`, or `type_of_extraction`).

- Update tests if needed.

viadot/tasks/sap_rfc.py
```diff
@@ -68,42 +73,69 @@ def __init__(
...
         if query is None:
             raise ValueError("Please provide the query.")
-        sap = SAPRFC(
-            sep=sep,
-            credentials=credentials,
-            func=func,
-            rfc_total_col_width_character_limit=rfc_total_col_width_character_limit,
-        )
+
+        if new_approx:
```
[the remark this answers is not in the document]
Diego-H-S: done!

Diego-H-S (commented):
Everything was done, let me know if you need more things to add.

Rafalz13 (approved):

PR #577, opened 2022-12-22
Check if dtypes is sorted as in ADLS file

Diego-H-S:
## Summary
Before adding this new feature, if the user passes a non-sorted dtypes argument like in the ADLS file, it creates an unsorted data frame to upload to Azure SQL table.
Also added more descriptions to azure_sql.py about how the credentials are loaded.

## Importance
Not relevant right now, because all our flows are well built but it would be important for any user that is not aware of the problem.

## Checklist

This PR:

### DIFF
--- a/tests/integration/flows/test_adls_to_azure_sql.py
+++ b/tests/integration/flows/test_adls_to_azure_sql.py
@@ -81,6 +81,7 @@ def test_check_dtypes_sort():
         "col2": "varchar(6)",
     }
     task = check_dtypes_sort
+
     n_dtypes = task.run(df=df, dtypes=dtypes)
     assert list(dtypes.keys()) == list(n_dtypes.keys())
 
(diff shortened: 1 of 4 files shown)


viadot/flows/adls_to_azure_sql.py
```diff
@@ -5,8 +5,10 @@
 import pandas as pd
 from prefect import Flow, task
 from prefect.backend import get_key_value
+from prefect.engine import signals
 from prefect.utilities import logging
 
+
```
Rafalz13, line 11: delete extra line

Rafalz13 (changes requested):
Please add tests to the new functionality.

tests/integration/flows/test_adls_to_azure_sql.py
```diff
@@ -67,3 +70,35 @@ def test_df_to_csv_task_none(caplog):
...
+    }
+    task = check_dtypes_sort
+    n_dtypes = task.run(df=df, dtypes=dtypes)
+    assert list(dtypes.keys()) == list(n_dtypes.keys())
+
+    dtypes = {
+        "col2": "varchar(6)",
+        "col1": "varchar(6)",
+    }
+    task = check_dtypes_sort
```
Rafalz13, line 91: Do you have to overwrite the task object every time? Maybe once on line 83, then just  run it.

Diego-H-S: You are right, changed!

Rafalz13 (approved):

PR #584, opened 2023-01-05
Sharepoint dl new module

cgildenia:
## Summary
Added a new download function for Sharepoint using a new module: Office365

## Importance
Due to the previous authentication method being deprecated, an alternative was needed

## Checklist

This PR:

### DIFF
--- a/viadot/sources/sharepoint.py
+++ b/viadot/sources/sharepoint.py
@@ -2,6 +2,10 @@ from typing import Any, Dict
 
 import sharepy
 
+import os
+from office365.sharepoint.files.file import File
+from office365.runtime.auth.user_credential import UserCredential
+
 from ..config import local_config
 from ..exceptions import CredentialError
 from .base import Source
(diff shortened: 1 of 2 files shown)


viadot/sources/sharepoint.py
```diff
@@ -65,3 +69,29 @@ def download_file(
             url=download_from_path,
             filename=download_to_path,
         )
+
+    def download_file_from_url(self, rel_file_url, download_folder: str = "./") -> None:
```
Rafalz13, line 73: Delete the previous method `download_file`. Add an option to provide a file name not only download_folder.
You get the filename on line 89, but it should be possible to provide that name from flow/task. Check the old method (lines 55-56)

viadot/sources/sharepoint.py
```diff
@@ -65,3 +69,29 @@ def download_file(
...
+            abs_url (str): URL to sharepoint website (e.g.: {tenant_name}.sharepoint.com)
+            rel_file_url (str): Relative URL to the file (e.g.: /sites/{directory}/Shared%20Documents/Downloads/)
+            credentials (str,str): Tuple with user and password (for now)
+            download_folder (str): Path where the file will be downloaded to
+
+        """
+        abs_url = self.credentials["site"]
+        rel_file_url = rel_file_url or self.url
+        user_name = self.credentials["username"]
+        user_password = self.credentials["password"]
```
Rafalz13, line 85: In the final version, it will be client_id and client_secret. It should be initialized in the `__init__` method not in the `doenload_...` method.

viadot/sources/sharepoint.py
```diff
@@ -65,3 +69,29 @@ def download_file(
...
+
+        abs_file_url = str(abs_url + rel_file_url)
```
Rafalz13, line 87: Maybe it will be easier to pass the full URL (The current solution uses this logic). It's easier for the user and we only need the full URL at the end.

Rafalz13 (commented):
Some comments added. Check if tests need to be updated after changes

PR #591, opened 2023-01-16
Mindful Not File Exception catch up

Diego-H-S:
## Summary
Implemented new logic when interactions and response files are not created in `Mindful`. With this new logic Prefect related to the CIC Mindful project will finish with success instead of failure as of now. We have created an empty file in ADLS to go there when those files are not created to be able to create an empty table.

## Importance
Required by client.

## Checklist

This PR:

### DIFF
--- a/viadot/flows/mindful_to_adls.py
+++ b/viadot/flows/mindful_to_adls.py
@@ -4,16 +4,20 @@ import pandas as pd
 
 from datetime import datetime
 from prefect import Flow, task
+from prefect.utilities import logging
+from prefect.engine.signals import FAIL
+from prefect.triggers import all_successful
 from viadot.tasks import MindfulToCSV
 from viadot.tasks import AzureDataLakeUpload
 from viadot.task_utils import add_ingestion_metadata_task
 
+logger = logging.get_logger()
 file_to_adls_task = AzureDataLakeUpload()
 
 
 @task
 def adls_bulk_upload(
-    file_names: List[str],
+    file_names: List[str] = None,
     file_name_relative_path: str = "",
     adls_file_path: str = None,
     adls_sp_credentials_secret: str = None,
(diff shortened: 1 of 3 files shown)


Rafalz13 (commented):
Please add tests for new code - if files/paths are not found.

viadot/flows/mindful_to_adls.py
```diff
@@ -33,14 +37,17 @@ def adls_bulk_upload(
...
-    for file in file_names:
-        file_path = str(adls_file_path + "/" + file)
-        file_to_adls_task.run(
-            from_path=os.path.join(file_name_relative_path, file),
-            to_path=file_path,
-            sp_credentials_secret=adls_sp_credentials_secret,
-            overwrite=adls_overwrite,
-        )
+    if not file_names:
+        logger.warning("Avoided uploading any file to ADLS. No files were reported.")
```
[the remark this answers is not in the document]
Diego-H-S: It's the same structure, but the intentions are completely different.
In the task class, I only do that to return `None` and to not let flow fails (that is what the client wants). But I put that in the Flow-task, because it is supposed that that task is going to become a general task in viadot. I think more convenient the warning there instead of in the task class.

tests/integration/test_mindful.py
```diff
@@ -89,3 +98,11 @@ def test_mindful_responses(mock_connection):
     assert mf.endpoint == "responses" and isinstance(mf.endpoint, str)
     assert os.path.exists("responses.csv")
     os.remove("responses.csv")
+
+
+@mock.patch("viadot.sources.Mindful._mindful_api_response", return_value=MockClass2)
+@pytest.mark.exception
+def test_file_exception(mock_nindful_1):
```
Rafalz13, line 105: Correct typo `.._nindful_..`

tests/integration/test_mindful.py
```diff
@@ -1,4 +1,5 @@
 import os
+import ast
```
Rafalz13, line 2: Do you need this import?

Rafalz13 (commented):
In the test `test_api_response` (line 50) you used a `viadot.sources.mindful.handle_api_response` but for the rest, you use `_mindful_api_response`. Should it be updated? This test has no `assert`.

tests/integration/test_mindful.py
Diego-H-S: no, removed.

tests/integration/test_mindful.py
Diego-H-S: done!

viadot/flows/mindful_to_adls.py
```diff
@@ -15,10 +19,13 @@ def add_timestamp(files_names: List = None, sep: str = "\t") -> None:
         files_names (List, optional): File names where to add the new column. Defaults to None.
         sep (str, optional): Separator type to load and to save data. Defaults to "\t".
     """
-    for file in files_names:
-        df = pd.read_csv(file, sep=sep)
-        df_updated = add_ingestion_metadata_task.run(df)
-        df_updated.to_csv(file, index=False, sep=sep)
+    if not files_names:
+        logger.warning("Avoided adding a timestamp. No files were reported.")
```
Rafalz13, line 23: Please create a `logger` object. You imported `logging` and then used `logger.`

Rafalz13 (commented):
Please correct `logger` part

PR #608, opened 2023-02-14
✨ Add `test_download_file()` test for `Sharepoint` source

AnnaGerlich:
## Summary

- Added missing test for `Sharepoint.download_file()`

## Checklist

This PR:

### DIFF
--- a/tests/unit/test_sharepoint.py
+++ b/tests/unit/test_sharepoint.py
@@ -1,3 +1,5 @@
+import os
+
 from viadot.sources import Sharepoint
 
 
@@ -5,3 +7,14 @@ def test_to_df(sharepoint_url):
     s = Sharepoint(config_key="sharepoint_prod")
     df = s.to_df(sharepoint_url)
     assert not df.empty
+
+
+def test_download_file(sharepoint_url):
+    s = Sharepoint(config_key="sharepoint_prod")
+
+    file = "sharepoint_test" + sharepoint_url.split(".")[-1]
+    s.download_file(url=sharepoint_url, to_path=file)
+
+    assert file in os.listdir()
+
+    os.remove(file)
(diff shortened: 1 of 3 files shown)


Trymzet:
Actually I don't think it makes sense to do `from viadot.sources import SharepointCredentials`; I would only expect sources to be importable that way. I think importing credentials with `from viadot.sources.sharepoint import SharepointCredentials` is more intuitive.

tests/unit/test_sharepoint.py
```diff
@@ -1,7 +1,20 @@
...
 
 
 def test_to_df(sharepoint_url):
     s = Sharepoint(config_key="sharepoint_prod")
     df = s.to_df(sharepoint_url)
     assert not df.empty
+
+
+def test_download_file(sharepoint_url):
+    s = Sharepoint(config_key="sharepoint_prod")
```
Trymzet, line 13: Let's find a way not to publicly declare we don't have a test environment. 😅 If there is actually no dev or test envs, maybe we can create a fixture for `Sharepoint` and then this config key can be taken from an env variable that we set just for testing, eg. `VIADOT_TEST_SHAREPOINT_CONFIG_KEY`?

Trymzet (changes requested):

tests/unit/test_sharepoint.py
AnnaGerlich: [✨ Added support for sharepoint_config_key](https://github.com/dyvenia/viadot/pull/608/commits/0d21f22d013bd32ad4801aabf6051cdf0a8ad7b3)

Trymzet (approved):
Looks good

Trymzet:
Pls rebase @AnnaGerlich then I can merge

Trymzet (approved):

Issue #586, opened 2023-01-13
Add remaining fixes from #579


Trymzet:
Add remaining fixes from #579

AnnaGerlich:
`s3fs.S3FileSystem()` doesn't support `region_name`

https://github.com/dyvenia/viadot/blob/084af16925343b3ec636205d6f39bc9f27497aed/viadot/sources/s3.py#L40

AnnaGerlich:
Remove `if` to get file type as it results with creating files called
`s3://bucket/folder/file.parquet/a734ad32537043af93ba747ab75c01fc.snappy.parquet`, which is not desired

https://github.com/dyvenia/viadot/blob/084af16925343b3ec636205d6f39bc9f27497aed/viadot/sources/s3.py#L135

AnnaGerlich:
`RedshiftSpectrum.rm()` doesn't work with `remove_files=True` > error occurs when removing S3 file

https://github.com/dyvenia/viadot/blob/084af16925343b3ec636205d6f39bc9f27497aed/viadot/sources/redshift_spectrum.py#L115

AnnaGerlich:
`redshift_source.rm()` doesn't remove redshift tables

AnnaGerlich:
`S3.exists()` doesn't work for directories

https://github.com/dyvenia/viadot/blob/084af16925343b3ec636205d6f39bc9f27497aed/viadot/sources/s3.py#L72

AnnaGerlich:
Add tests to `S3` and `RedshiftSpectrum` sources

PR #579, opened 2022-12-27
Add s3 and redshift_spectrum sources

AnnaGerlich:
## Summary

- added s3 source
- added spectrum source
- added boto3 to the requirements
- added awswrangler to the requirements

## Importance

Basic AWS sources to transfer data

PR #636, opened 2023-03-02
Added tests for the `RedshiftSpectrum` source

AnnaGerlich:
## Summary

- Added `RedshiftSpectrumCredentials` class 
- Renamed `rm()` method to `drop_table()`
- Fix session handling in `drop_table()` method
- Added test for RedshiftSpectrum source
- Bumped viadot base image version to `2.0-latest`

## Importance

- Test for RedshiftSpectrum source was previously missing
- There will be no need to update viadot image tag as the `2.0-latest` is fixed
- `RedshiftSpectrum.rm()` method will be now called `RedshiftSpectrum.drop_table()`, which is more user friendly and will now also work when providing `config_key` (which previously was triggering an error).

## Issue 
[Add remaining fixes from #579](https://github.com/dyvenia/viadot/issues/586)

## Checklist

This PR:


tests/conftest.py
```diff
@@ -82,8 +82,8 @@ def TEST_ADLS_FILE_PATH_CSV():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def s3_config_key():
-    return os.environ.get("S3_CONFIG_KEY")
+def aws_config_key():
+    return os.environ.get("AWS_CONFIG_KEY")
```
Trymzet, line 86: Pls then also modify relevant `.env.example` (the one in `tests/`?)

viadot/sources/redshift_spectrum.py
```diff
@@ -79,8 +87,8 @@ def exists(self, database: str, table: str) -> bool:
         Check if a table exists in Glue database.
 
         Args:
-            database (str): AWS Glue catalog database name.
-            table (str): AWS Glue catalog table name.
+            database (str): Amazon Glue catalog database name.
+            table (str): Amazon Glue catalog table name.
```
Trymzet, line 91: It's actually called AWS Glue, remember that Amazon is an e-commerce company, AWS is its cloud division. I think Redshift is one of the exceptions when they use the Amazon name, probably because it's quite old?

tests/unit/test_redshift_spectrum.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    redshift.drop_table(database=TEST_SCHEMA, table=TEST_TABLE, remove_files=True)
+
+    assert result is True
+
+
+def test_to_df(redshift):
+    redshift.from_df(
+        df=TEST_DF,
+        to_path=f"s3://datawerfen-159170848751291/nesso/{TEST_SCHEMA}/{TEST_TABLE}",
```
Trymzet, line 56: Remove client info

tests/unit/test_redshift_spectrum.py
```diff
@@ -0,0 +1,68 @@
...
+        database=TEST_SCHEMA,
+        table=TEST_TABLE,
+    )
+
+    result = redshift.exists(
+        database=TEST_SCHEMA,
+        table=TEST_TABLE,
+    )
```
Trymzet, line 47: Can you pls add a test for `exists()` before this one? Otherwise this test is not very trustworthy 😅

tests/unit/test_redshift_spectrum.py
```diff
@@ -0,0 +1,68 @@
...
+
+    result = redshift.to_df(
+        database=TEST_SCHEMA,
+        table=TEST_TABLE,
+    )
+
+    assert len(result) == len(TEST_DF)
```
Trymzet, line 66: We can just check for equality here? Like `df.equals(other_df)`

Trymzet (changes requested):

tests/conftest.py
AnnaGerlich: [🚚 Rename S3_CONFIG_KEY env var to AWS_CONFIG_KEY](https://github.com/dyvenia/viadot/pull/636/commits/7719be2cae187395bf54db858fff32e504c387ea)

viadot/sources/redshift_spectrum.py
AnnaGerlich: Agreed. I've reverted previous commit and changes AWS S3 > Amazon S3 as per [📝 Minor changes](https://github.com/dyvenia/viadot/pull/636/commits/f8c1fcefb096de9daeb1785594afb86cd9a111d0)

tests/unit/test_redshift_spectrum.py
AnnaGerlich: [🐛 Fix bucket name to S3_BUCKET env var](https://github.com/dyvenia/viadot/pull/636/commits/8859163306b55966c4daca09d0f5467a0670e3af)

tests/unit/test_redshift_spectrum.py
AnnaGerlich: [⚡️ Improve test assertion](https://github.com/dyvenia/viadot/pull/636/commits/2916fdaf3b1ae810b28ed9375f9a3c6e5c0953bd)

tests/unit/test_redshift_spectrum.py
AnnaGerlich: [⚡️ Improve test assertion](https://github.com/dyvenia/viadot/pull/636/commits/2916fdaf3b1ae810b28ed9375f9a3c6e5c0953bd)

tests/unit/test_redshift_spectrum.py
AnnaGerlich: [✅ Add test for RedshiftSpectrum.exists() method](https://github.com/dyvenia/viadot/pull/636/commits/bd0e4e356110595645717db6c65ddf54f467ccf4)

Trymzet: Ah sry, I think you should also run exists() before running from_df(), so, asserst exist is False -> create table -> assert exists is True, otherwise this table could have been there all along, even before you ran `from_df()`. Also this exists test, the way it is right now... does that make sense? Does it *really* test something? :p

Trymzet (changes requested):

PR #638, opened 2023-03-06
Fixed bug in exchange rates unit tests

djagoda881:
## Summary

The following PR fixes a bug in one of the tests in the ExchangeRates source. This is exactly the test `test_to_df_values`. The problem was caused by the fact that when the to_df() method is called on the object, 2 additional columns are added, so the data was not identical to the expected ones.

## Importance

After the fix all tests will work.

## Checklist

This PR:

### DIFF
--- a/tests/unit/test_exchange_rates.py
+++ b/tests/unit/test_exchange_rates.py
@@ -59,25 +59,26 @@ def test_to_json_values():
     assert retrieved_value == expected_value
 
 
-# def test_to_df_values():
-#     source = ExchangeRates(
-#         currency="PLN",
-#         start_date="2022-10-09",
-#         end_date="2022-10-11",
-#         symbols=["USD", "EUR", "GBP", "CHF", "PLN", "DKK"],
-#         config_key="exchange_rates_dev",
-#     )
-#     expected_value = TEST_DF
-#     retrieved_value = source.to_df()
-#     retrieved_value.drop(["_viadot_downloaded_at_utc"], axis=1, inplace=True)
-#     assert retrieved_value.iloc[0].equals(expected_value.iloc[0])
+def test_to_df_values():
+    source = ExchangeRates(
+        currency="PLN",
+        start_date="2022-10-09",
+        end_date="2022-10-11",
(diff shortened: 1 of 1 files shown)


tests/unit/test_exchange_rates.py
```diff
@@ -59,25 +59,26 @@ def test_to_json_values():
...
+        currency="PLN",
+        start_date="2022-10-09",
+        end_date="2022-10-11",
+        symbols=["USD", "EUR", "GBP", "CHF", "PLN", "DKK"],
+        config_key="exchange_rates_dev",
+    )
+    expected_value = TEST_DF
+    retrieved_value = source.to_df()
+    retrieved_value.drop(["_viadot_source"], axis=1, inplace=True)
+    retrieved_value.drop(["_viadot_downloaded_at_utc"], axis=1, inplace=True)
```
Trymzet, line 73: I think this will break any time we change the util func that adds these columns, right? So it would be better if you somehow retrieve the columns to be removed from that util function, or maybe there's a variable like `VIADOT_MATADATA_COLS` or sth. BTW I think you can do the drops in one line, `.drop(metadata_cols)`.

Trymzet (changes requested):

tests/unit/test_exchange_rates.py
Trymzet: But you didn't fix the problem that any time the util changes, your unrelated tests will break. This is why you should import the metadata cols from `utils.py`, so this test reuses the same columns as `add_viadot_metadata_columns()`.

Trymzet (changes requested):
Same issue as last time as it wasn't fixed.

tests/unit/test_exchange_rates.py
djagoda881: I have improved the way the to_df function is called. The function is now run without a decorator. This way it only tests the correct function in the ExchangeRates class

[✅ Updated test_to_df_values](https://github.com/dyvenia/viadot/pull/638/commits/6ecd82fa93827f613a4436bd6c4ed8ca61ef353a)

PR #652, opened 2023-03-13
Genesys new conversations endpoint

Diego-H-S:
## Summary
A new endpoint connection is required by the client to Genesys API. This new "conversations details query" endpoint requires some modifications in Genesys task and flow, however. A new method `merge_conversations_dfs` in the task file is created to manage an extremely nested JSON response. In the flow level, a new function `filter_userid` is added to return only conversations for a list of user IDs passed as an argument by the final user. Minor changes were introduced in the Source file to return the response in this new endpoint.

## Importance
Very important. Required by the client.

## Checklist

This PR:


viadot/flows/genesys_to_adls.py
```diff
@@ -26,13 +31,57 @@ def add_timestamp(files_names: List = None, path: str = "", sep: str = "\t") ->
...
+    Args:
+        files_names (list, optional): All file names of downloaded files. Defaults to "\t".
+        path (str, optional): Relative path to the file. Defaults to empty string.
+        userids (list, optional): List of all user IDs to select in the data frame. Defaults to None.
+        sep (str, optional): Separator in csv file. Defaults to None.
+        apply_method (bool, optional): Use this method or avoid its execution. Defaults to False.
+    """
+
+    if apply_method:
+        for file in files_names:
```
Rafalz13, line 53: `files_names` is  `None` as default. What if there file_names will be None? In this case, we will receive an error that 'Nonetype' is not iterable. Maybe add some `try-except block` here.

viadot/flows/genesys_to_adls.py
```diff
@@ -26,13 +31,57 @@ def add_timestamp(files_names: List = None, path: str = "", sep: str = "\t") ->
...
+            df = pd.read_csv(os.path.join(path, file), sep=sep)
+
+            # first: it gets all the conversations ID where an agent is present.
+            conversations_id = np.array([])
+            for user in userids:
```
Rafalz13, line 58: The same situation as above. What if user_ids is `None`?

viadot/flows/genesys_to_adls.py
```diff
@@ -26,13 +31,57 @@ def add_timestamp(files_names: List = None, path: str = "", sep: str = "\t") ->
...
+    userids: list = None,
+    apply_method: bool = False,
+) -> None:
+    """filter out the data frame by user ID.
+
+    Args:
+        files_names (list, optional): All file names of downloaded files. Defaults to "\t".
+        path (str, optional): Relative path to the file. Defaults to empty string.
+        userids (list, optional): List of all user IDs to select in the data frame. Defaults to None.
+        sep (str, optional): Separator in csv file. Defaults to None.
```
Rafalz13, line 48: Just a note - maybe change `sep` to `wite_sep` or modify the docstring to inform that this is the separator for the CSV file that will be created from this function.

viadot/sources/genesys.py
```diff
@@ -190,6 +194,9 @@ async def generate_post():
         coroutine = generate_post()
         loop.run_until_complete(coroutine)
 
+        if end_point == "conversations/details/query":
+            return json.loads(new_report.decode("utf-8"))
```
Rafalz13, line 198: The function returns 'dict' only if `endpoint = conversations/details/query`, right? What if the `end_point` is different?

viadot/tasks/genesys.py
```diff
@@ -74,6 +82,133 @@ def __call__(self, *args, **kwargs):
         """Download Genesys data to CSV"""
         return super().__call__(*args, **kwargs)
 
+    def merge_conversations_dfs(self, data_to_merge: list) -> DataFrame:
+        """Method to merge all the conversations data into a single data frame.
```
Rafalz13, line 86: Is it possible to describe what is happening inside this function? You have a lot of `normilize` and FORs. I will be easier to understand.

viadot/tasks/genesys.py
```diff
@@ -166,33 +305,86 @@ def run(
...
 
-        if "FAILED" in failed and "COMPLETED" in failed:
-            logger.warning("Some reports failed.")
+            if "FAILED" in failed and "COMPLETED" in failed:
+                logger.warning("Some reports failed.")
 
-        if len(genesys.report_data) == 0 or np.unique(failed)[0] == "FAILED":
+            if len(genesys.report_data) == 0 or np.unique(failed)[0] == "FAILED":
+                genesys.delete_all_reporting_exports()
+                logger.warning(f"All existing reports were delted.")
```
Rafalz13, line 326: `delted`? - probably a typo

viadot/tasks/genesys.py
```diff
@@ -166,33 +305,86 @@ def run(
...
+            )
+            logger.info("Downloaded the data from the Genesys into the CSV.")
+            # in order to wait for API GET request call it
+            logger.info("Waiting for caching data in Genesys database.")
+            time.sleep(20)
             genesys.delete_all_reporting_exports()
-            logger.warning(f"All existing reports were delted.")
-            raise APIError("No exporting reports were generated.")
-        elif not None in [col for col in np.array(genesys.report_data).T][1]:
+            logger.info(f"All existing reports were delted.")
```
Rafalz13, line 341: `delted` - typo?

viadot/tasks/genesys.py
```diff
@@ -166,33 +305,86 @@ def run(
...
-        logger.info("Downloaded the data from the Genesys into the CSV.")
-        # in order to wait for API GET request call it
-        logger.info("Waiting for caching data in Genesys database.")
-        time.sleep(20)
-        genesys.delete_all_reporting_exports()
-        logger.info(f"All existing reports were delted.")
-
-        return file_names
+
+            return [file_name]
```
Rafalz13, line 390: `return [file_name]` because here we have only 1 file, right? above you have `return file_names`

Rafalz13 (changes requested):
Added few comments.
Please test the rest of Genesys flows using new code.
There are probably a few places where you can replace the code block with a function, but for now please apply the changes.

viadot/flows/genesys_to_adls.py
```diff
@@ -1,78 +1,133 @@
...
         df_updated = add_ingestion_metadata_task.run(df)
-        df_updated.to_csv(file, index=False, sep=sep)
+        df_updated.to_csv(os.path.join(path, file), index=False, sep=sep)
+
+
+@task(timeout=3600)
+def filter_userid(
+    files_names: list = None, path: str = "", sep: str = "\t", userids: list = None
+) -> None:
+    """filter out the data frame by user ID.
```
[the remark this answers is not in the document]
Diego-H-S: done!

viadot/flows/genesys_to_adls.py
```diff
@@ -1,78 +1,133 @@
...
-        df = pd.read_csv(file, sep=sep)
+        df = pd.read_csv(os.path.join(path, file), sep=sep)
         df_updated = add_ingestion_metadata_task.run(df)
-        df_updated.to_csv(file, index=False, sep=sep)
+        df_updated.to_csv(os.path.join(path, file), index=False, sep=sep)
+
+
+@task(timeout=3600)
+def filter_userid(
+    files_names: list = None, path: str = "", sep: str = "\t", userids: list = None
```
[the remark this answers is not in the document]
Diego-H-S: done!

viadot/flows/genesys_to_adls.py
```diff
@@ -26,13 +31,57 @@ def add_timestamp(files_names: List = None, path: str = "", sep: str = "\t") ->
...
+        sep (str, optional): Separator in csv file. Defaults to None.
+        apply_method (bool, optional): Use this method or avoid its execution. Defaults to False.
+    """
+
+    if apply_method:
+        for file in files_names:
+            df = pd.read_csv(os.path.join(path, file), sep=sep)
+
+            # first: it gets all the conversations ID where an agent is present.
+            conversations_id = np.array([])
```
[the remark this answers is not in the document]
Diego-H-S: yes because I use a numpy method named `np.unique()` very useful, otherwise I have to do it by myself in a new function when It is already done and probably more efficient.

viadot/flows/genesys_to_adls.py
```diff
@@ -26,13 +31,57 @@ def add_timestamp(files_names: List = None, path: str = "", sep: str = "\t") ->
         df_updated.to_csv(os.path.join(path, file), index=False, sep=sep)
 
 
+@task(timeout=3600)
+def filter_userid(
+    files_names: list = None,
```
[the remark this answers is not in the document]
Diego-H-S: done!

viadot/flows/genesys_to_adls.py
```diff
@@ -26,13 +31,57 @@ def add_timestamp(files_names: List = None, path: str = "", sep: str = "\t") ->
...
+    path: str = "",
+    sep: str = "\t",
+    userids: list = None,
```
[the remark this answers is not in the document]
Diego-H-S: done!

viadot/sources/genesys.py
Diego-H-S: Returns None, that's why I use the typing `Union[None, dict]`. is that why you are asking for it?
I updated the doc-string.

viadot/tasks/genesys.py
Diego-H-S: I add and example and some more comments in the code, have a look at it!

viadot/tasks/genesys.py
Diego-H-S: yes, we only get one file and the flow is expecting a list of files because the other endpoint could return a list of returned files.

viadot/flows/genesys_to_adls.py
Diego-H-S: you are right, I added an `if` checking to avoid that problem. I always avoid using try-except statements to control the flow.

viadot/flows/genesys_to_adls.py
Diego-H-S: Added this possibility to an `if` statement.

viadot/sources/genesys.py
```diff
@@ -160,6 +160,9 @@ def genesys_generate_exports(
         Args:
             post_data_list (List[str], optional): List of string templates to generate json body. Defaults to None.
             end_point (str, optional): Final end point for Genesys connection. Defaults to "reporting/exports".
+
+        Returns:
+            Union[None, dict]: Dict when the "conversations" endpoint is called, otherwise returns None.
```
Rafalz13, line 165: I cannot find the second `return`. You wrote that it should be `Dict` or `None`

Rafalz13 (commented):
Looks good. I added one additional comment.

viadot/sources/genesys.py
Diego-H-S: changed by Optional, as it is said in python documentation, it should be the best practice.

Rafalz13 (approved):

viadot/flows/genesys_to_adls.py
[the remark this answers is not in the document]
Rafalz13: For getting unique values you can use `set()` - I'm not sure if this is a good solution in this case, but if so, you can use a solution like this

PR #660, opened 2023-04-04
🚀Add dev_customer_gauge to viadot:dev

hhnks:
## Summary
Source class `CustomerGauge` connects and downloads data using Customer Gauge API. `CustomerGauge` source can connect to one of the selected endpoint: *responses* and *non-responses*. Authentication to the API is done via Bearer Token. The `CustomerGauge` class contains : 
- `get_token` function that gets Bearer Token using POST request method
- `get_json_response` function that gets JSON with nested structure that contains data and cursor parameter value using GET request method
- `get_cursor` function which returns the cursor value that is needed to navigate to the next page in the next API call for specific pagesize
- `properties_cleaning` function that returns initialy cleaned data. The cleaning of the additional params is depend on the endpoint
- `flatten_json` function that flattens a nested structure of the JSON object into a single-level dictionary.
        Uses a nested `flatten()` function to recursively combine nested keys in the JSON object with '_' to create the flattened keys.
- `to_df` function that combines all of the functions above and converts JSON data it into flat pandas DataFrame with cleaned column names and it's data

Task class `CustomerGaugeToDF` downloads the selected range of data from the selected Customer Gauge endpoint and returns as one pandas DataFrame. Since main limitation of API is retriving maximum 1000 rows at once, task class uses built-in while loop to download range of data selected by user.

Flow class `CustomerGaugeToADLS` downloads data from the Customer Gauge's endpoints (Responses and Non-Responses) via API to a CSV or Parquet file and upload it to Azure Data Lake. In a flow class, there is possible data anonymization if needed by user. The anonymization is handled using `anonymize_df` from `task_utils` and it happens before uploading data do Azure Data Lake.

## Importance
The connector enables automated and scheduled data flows from Customer Gauge. 

## Checklist

This PR:


viadot/tasks/customer_gauge.py
```diff
@@ -0,0 +1,146 @@
...
+                    f"Downloading starting from the {cursor} cursor. Process might take a few minutes..."
+                )
+            while df.empty == False:
+                json_data = customer_gauge.get_json_response(cursor=cur)
+                cur = customer_gauge.get_cursor(json_data)
+                df = customer_gauge.to_df(json_data)
+                df_list.append(df)
+
+        df_total = pd.concat(df_list, ignore_index=True)
+        logger.info(f"Downloading finished.")
```
Rafalz13, line 144: I am not sure if this logger is necessary here because at the end after the task is finished in logger, we can see the message that `Finished task run for task with final state: 'Success'`

viadot/tasks/customer_gauge.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Returns:
+            pd.DataFrame: Final pandas DataFrame.
+        """
+        try:
+            credentials_str = AzureKeyVaultSecret(
+                credentials_secret, vault_name=vault_name
+            ).run()
+        except ValueError:
+            pass
```
Rafalz13, line 110: Add a message: "Credentials not found." or something similar -  If it's not logged anywhere in the task.

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+            else:
+                raise ValueError(
+                    "Incorrect endpoint name. Choose: 'responses' or 'non-responses'"
+                )
+        else:
+            raise ValueError(
+                "Provide endpoint name. Choose: 'responses' or 'non-responses'"
+            )
+
+        if credentials is not None:
```
Rafalz13, line 49: Run super().__init__(credentials=credentials) - (check other sources)  after`if credentials.....` block. After that, you can refer to credentials with `self.`

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+            self.credentials = credentials
```
Rafalz13, line 50: You can also add an option with local_config - import local config. Then you can replace `credentials=credentials` with `credentials= DEFAULT_CREDENTIALS or credentials`

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+            "client_id": client_id,
+            "client_secret": client_secret,
+        }
+        api_response = handle_api_response(
+            url=url, params=body, headers=headers, method="POST"
+        )
+        token = api_response.json().get("access_token")
+
+        if token is None:
+            raise APIError("Token is None. Check your credentials.")
```
Rafalz13, line 82: Change the message e.g( `The token could not be generated. Check your credentials`)

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+
+        try:
+            cur = json_response["cursor"]["next"]
+        except:
+            raise ValueError(
+                "Provided argument doesn't contain 'cursor' value. Pass json returned from the endpoint."
+            )
+
+        return cur
+
```
Rafalz13, line 164: Between `return` and `def new_function` should be 2 empty lines. Please correct in the entire file.

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+        flattify(json_response)
+
+        return out
+
+    def to_df(self, json_response: Dict[str, Any] = None) -> pd.DataFrame:
+        """
+        Flatten dictionary structure and convert it into pandas DataFrame. Clean column names
+
+        Args:
+            json_response (Dict[str, Any], optional): JSON object represented as a nested dictionary that contains data and cursor parameter value.. Defaults to None.
```
Rafalz13, line 232: Remove one dot -> `value.. Defaults`

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+            else:
+                out[key[:-1]] = x
+
+        flattify(json_response)
+
+        return out
+
+    def to_df(self, json_response: Dict[str, Any] = None) -> pd.DataFrame:
+        """
+        Flatten dictionary structure and convert it into pandas DataFrame. Clean column names
```
Rafalz13, line 229: Add full stop.

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        Gets JSON with nested structure that contains data and cursor parameter value using GET request method.
+
+        Args:
+            url (str, optional): Endpoint URL. Defaults to None.
+            cursor (int, optional): Cursor value to navigate to the page. Defaults to None.
+            pagesize (int, optional): Number of responses (records) returned per page, max value = 1000. Defaults to 1000. Defaults to 1000.
+            date_field (Literal["date_creation", "date_order", "date_sent", "date_survey_response"], optional): Specifies the date type which filter date range. Defaults to None.
+            start_date (datetime, optional): Defines the period start date in yyyy-mm-dd format. Defaults to None.
+            end_date (datetime, optional): Defines the period end date in yyyy-mm-dd format.. Defaults to None.
```
Rafalz13, line 106: Remove one dot -> `format.. Defaults`

viadot/sources/customer_gauge.py
```diff
@@ -0,0 +1,250 @@
...
+            credentials (Dict[str, Any], optional): Credentials to connect with API containing client_id, client_secret. Defaults to None.
+
+        Raises:
+            ValueError: If endpoint is not provided or incorect.
+            CredentialError: If credentials are not provided in local_config or directly as a parameter
+        """
+        self.endpoint = endpoint
+        if endpoint is not None:
+            if endpoint in ["responses", "non-responses"]:
+                self.url = f"{self.API_URL}{endpoint}"
```
Rafalz13, line 39: You set API_URL as a global variable. Is it correct to call `self.API_URL`?

Rafalz13 (changes requested):
Some comments added.

viadot/sources/customer_gauge.py
hhnks: what do you mean?

viadot/tasks/customer_gauge.py
hhnks: right, now it looks like that:
image


I will remove this one

viadot/tasks/customer_gauge.py
hhnks: I'm not sure about that. If the credentials are not found in KV, then it checks local_config file. It's defined in source class.

viadot/sources/customer_gauge.py
hhnks: In this source, I don't use the base class (Source) and I included checking local_config instead (51-54 cells). 
Is it ok or should I remove this part and use inheritance from Source to keep consistency?

viadot/sources/customer_gauge.py
hhnks: as in conversation above

viadot/sources/customer_gauge.py
hhnks: In VSCode after saving it removed extra line I'm adding 😖

viadot/sources/customer_gauge.py
hhnks: I think keeping it as a class attribute helps to encapsulate it within the class, preventing it from being accessed or modified by other parts of the code and making it clear that this URL is associated with the class.

viadot/sources/customer_gauge.py
Rafalz13: Please add inheritance to have the same structure in all `sources` classes

viadot/tasks/customer_gauge.py
Rafalz13: It is ok if the error message is in the source.
In line 111 you have `credentials = json.loads(credentials_str)` if `credentials_str` is not found (=None) then this operation will return `TypeError: the JSON object must be str, bytes or bytearray, not NoneType`.

Please consider adding line 111 to `try except` block above. Then you can catch the `ValueError and TypeError`

viadot/sources/customer_gauge.py
Rafalz13: The message is still the same.

Rafalz13 (commented):
Added some comments to existing conversations. 
Tests looks good.

Rafalz13 (approved):
Looks good

PR #661, opened 2023-04-05
Utils response function

djagoda881:
## Summary

I replaced the `get_respone` function in utils with `handle_api_request` function. The reason for the change was to handle `POST` operations in the `genesys` source. The function is now generalised, accepting the `GET` parameter by default, so the change should not cause an error in other sources previously using `get_response`. 

## Checklist

This PR:


Trymzet:
@djagoda881 I don't see anything outside of `utils.py` changed here, so if this func is not used anywhere, why change it?

djagoda881:
The `handle_api_request` function is only called inside the `handle_api_response` function. The `cloud_for_customers` source uses this function and it will be used in the source that I am migrating which is `genesys`.
image

tests/unit/test_utils.py
```diff
@@ -68,6 +73,30 @@ def test_double_quotes_inside():
...
+    item = {
+        "name": "test_item",
+        "data": {"color": "blue", "price": 135},
+    }
+    payload = json.dumps(item)
+
+    response_post = handle_api_request(
+        url=url, method="POST", headers=headers, data=payload
+    )
+    assert response_post.status_code == 200
```
Trymzet, line 89: tip: checking for `response.ok` is more generic than for code 200 as it returns `True` for all 2xx codes.

viadot/utils.py
```diff
@@ -83,29 +109,37 @@ def handle_api_response(
...
     codes.
 
     Args:
-        url (str): the URL which trying to connect.
-        auth (tuple, optional): authorization information. Defaults to None.
-        params (Dict[str, Any], optional): the request parameters. Defaults to None.
-        headers: (Dict[str, Any], optional): the request headers. Defaults to None.
-        timeout (tuple, optional): the request times out. Defaults to (3.05, 60 * 30).
+        url (str): The URL which trying to connect.
+        auth (tuple, optional): Authorization information. Defaults to None.
```
Trymzet, line 121: 1) typo: it's authentication, not authorization
2) I would link here to request docs as it's unclear here what input is expected. See https://requests.readthedocs.io/en/latest/user/authentication/ for all possible auth types. As you can see, it's not just `tuple` (basic auth), but it can by any class that handles auth, some of which are provided by `requests`. And if we only support basic auth, this should be stated here, probably with an example (`(username, password)`).

viadot/utils.py
```diff
@@ -83,29 +109,37 @@ def handle_api_response(
...
+        params (Dict[str, Any], optional): The request parameters. Defaults to None.
+        headers: (Dict[str, Any], optional): The request headers. Defaults to None.
+        timeout (tuple, optional): The request times out. Defaults to (3.05, 60 * 30).
```
Trymzet, line 124: typo: times out->timeout. Also the original description is better as it explains what each value is: `A tuple of (connect_timeout, read_timeout) in seconds.`

viadot/utils.py
```diff
@@ -83,29 +109,37 @@ def handle_api_response(
...
 
     Raises:
-        ReadTimeout: stop waiting for a response after `timeout` seconds.
+        ReadTimeout: Stop waiting for a response after `timeout` seconds.
```
Trymzet, line 127: Timeout is a tuple, so this description is ambiguous. Timeout could be both 1 and 99999 seconds at the same time. But I think this should be ok if you fix `timeout` param docstring to explain which value is which.

viadot/utils.py
```diff
@@ -83,29 +109,37 @@ def handle_api_response(
...
         HTTPError: The raised HTTP error.
         ConnectionError: Raised when the client is unable to connect to the server.
-        APIError: viadot's generic API error.
+        APIError: Viadot's generic API error.
```
Trymzet, line 130: we use lowercase for `viadot`

Trymzet (changes requested):

viadot/utils.py
djagoda881: [📝 Updated docstrings in utils.py](https://github.com/dyvenia/viadot/pull/661/commits/ec9e1879f7155aa71233f9fad2aeba5560f290c0)

viadot/utils.py
djagoda881: [📝 Updated docstrings in utils.py](https://github.com/dyvenia/viadot/pull/661/commits/ec9e1879f7155aa71233f9fad2aeba5560f290c0)

viadot/utils.py
djagoda881: [📝 Updated docstrings in utils.py](https://github.com/dyvenia/viadot/pull/661/commits/ec9e1879f7155aa71233f9fad2aeba5560f290c0)

tests/unit/test_utils.py
djagoda881: [✅ Changed succesful status codes in api test](https://github.com/dyvenia/viadot/pull/661/commits/95b2c37ee28c8746d3ca5c5e448a9ac04c37239f)

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added rollback feature to `Databricks` source.
 - Changed all Prefect logging instances in the `sources` directory to native Python logging.
 - Changed `rm()`, `from_df()`, `to_df()` methods in `S3` Source
+- Changed `get_request function` to `handle_api_request`.
```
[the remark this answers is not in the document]
djagoda881: [📝 Corrected entry in CHANGELOG.md](https://github.com/dyvenia/viadot/pull/661/commits/36dc2fa118b007f3831ee861cd424e1a0aa229c9)

djagoda881:
Reverted two commits containing changes to the genesys.py file. They were accidentally added
[Revert "✨ Migrated genesys source from viadot 1"](https://github.com/dyvenia/viadot/pull/661/commits/786f341a87cd603228353b0c7e7b204e6594cdca)
[Revert "♻️ Refactored dostrings"](https://github.com/dyvenia/viadot/pull/661/commits/aaa4531c7aee0cb13efa2130113e288afeaeb153)

tests/unit/test_utils.py
Trymzet: Why not just use `response.ok`? It's already built into `requests`.

Trymzet (approved):

Issue #672, opened 2023-04-19
Databricks source does not use `df_snakecase_column_names()` function with `if_exists=replace` parameter.


djagoda881:
If the column names in the DataFrame that we send to Databricks use forbidden characters they will be changed by the `df_snakecase_column_names()` function before being sent to databricks. However, if the table already exists in the databricks and we want to overwrite it the `df_snakecase_column_names()` function is no longer called which causes an error because the columns have not been converted and cannot be sent to Databricks with forbidden characters.

PR #673, opened 2023-04-21
Databricks snakecase column bug

djagoda881:
## Summary

The issue (#672 ) has been fixed. The bug prevented overwriting a table in Databricks whose column names were not saved in "snake_case".  The fix was made by adding a call to the function `df_snakecase_column_names()` on the DataFrame inside the `_full_refresh()` function, which is executed when the `replace="True"` parameter is set.

## Importance

This is important because this bug could cause errors in workflows in the future.

## Checklist

This PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `RedshiftSpectrum` source to the library.
 - Added `upload()` and `download()` methods to `S3` source.
 - Added `Genesys` source to library.
-- Fixed bug in `Databricks._full_refresh()`.The function replacing column names with snake_case was not used. (#672)
+- Fixed a bug in `Databricks.create_table_from_pandas()`. The function that converts column names to snake_case was not used in every case. (#672)
 
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `RedshiftSpectrum` source to the library.
 - Added `upload()` and `download()` methods to `S3` source.
 - Added `Genesys` source to library.
+- Fixed bug in `Databricks._full_refresh()`.The function replacing column names with snake_case was not used. (#672)
+
 
 ### Changed
 - Added `SQLServerToDF` task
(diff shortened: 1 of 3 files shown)


tests/unit/test_databricks.py
```diff
@@ -233,6 +233,44 @@ def test_replace_different_column_schema(databricks):
     databricks.drop_schema(TEST_SCHEMA)
 
 
+def test_snakecase_column_names(databricks):
+
+    assert not databricks._check_if_table_exists(schema=TEST_SCHEMA, table=TEST_TABLE)
+
+    origin_to_df = databricks.to_df.__wrapped__
```
Trymzet, line 240: 1) origin is a noun that means something else than original;
2) it's not possible to know what "original" would mean here or why you're doing this instead of just using `to_df()`; this begs a comment and possibly a more explicit variable name, such as `to_df_no_metadata_cols()`.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `RedshiftSpectrum` source to the library.
 - Added `upload()` and `download()` methods to `S3` source.
 - Added `Genesys` source to library.
+- Fixed bug in `Databricks._full_refresh()`.The function replacing column names with snake_case was not used. (#672)
```
Trymzet, line 19: "fixed bug" -> "fixed a bug"; missing space after dot. You can use eg. Code Spell Checker VSCode extension to avoid doing these kind of mistakes in the future

Trymzet:
@djagoda881
1) I don't think this behavior should be hardcoded in `_full_refresh()`; it seems you can just move the line in `create_table_from_pandas()` above the if/else block.
2) Pls add a param in `create_table_from_pandas()` such as `snakecase_col_names: bool = True` to control this behavior

Trymzet (changes requested):
Added my comments

CHANGELOG.md
djagoda881: [📝 Improved entry in CHANGELOG.md](https://github.com/dyvenia/viadot/pull/673/commits/a78bc7ec3f216ba8efd6f9b3a30708eaa6146b2f)

tests/unit/test_databricks.py
djagoda881: [✅ Updated function name](https://github.com/dyvenia/viadot/pull/673/commits/4d28f6c1f51e2b9757186f6d3310cdca43a5d552)
[✅ Added comment to variable](https://github.com/dyvenia/viadot/pull/673/commits/250576c92b9a4072a7e6378d00a7419b55bbf9ec)

djagoda881:
> @djagoda881
> 
> 1. I don't think this behavior should be hardcoded in `_full_refresh()`; it seems you can just move the line in `create_table_from_pandas()` above the if/else block.
> 2. Pls add a param in `create_table_from_pandas()` such as `snakecase_col_names: bool = True` to control this behavior. There you can also document the reasoning for doing that as it's not clear from the code itself.

I corrected it as mentioned above.
[🎨 Added param to create_table_from_pandas function](https://github.com/dyvenia/viadot/pull/673/commits/bc58cb077fbfb3cafd8b4e6986b7e0308fd81316)

Trymzet (approved):
Looks good

PR #678, opened 2023-04-27
⚡️Enhanced `S3()` and `RedshiftSpectrum()` sources

AnnaGerlich:
## Summary

- Changed function to list tables
- Moved optional parameters to kwargs to allow more flexibility
- Reviewed dosctrings and descriptions

## Importance

- Old listing function `awswrangler.catalog.tables()` lists only 100 tables.
- Optional parameters were not correctly marked

## Checklist

This PR:

### DIFF
--- a/viadot/sources/redshift_spectrum.py
+++ b/viadot/sources/redshift_spectrum.py
@@ -135,31 +135,31 @@ class RedshiftSpectrum(Source):
         extension: str = ".parquet",
         if_exists: Literal["overwrite", "overwrite_partitions", "append"] = "overwrite",
         partition_cols: List[str] = None,
-        index: bool = False,
-        compression: str = None,
         sep: str = ",",
         description: str = None,
+        **kwargs,
     ) -> None:
         """
         Upload a pandas `DataFrame` to a csv or parquet file in Amazon Redshift Spectrum.
+            For full list of available parameters please refer to the official documentation:
+            https://aws-sdk-pandas.readthedocs.io/en/3.0.0/stubs/awswrangler.s3.to_parquet.html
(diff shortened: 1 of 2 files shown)


viadot/sources/redshift_spectrum.py
```diff
@@ -59,36 +59,32 @@ def session(self) -> boto3.session.Session:
...
     def ls(
         self,
-        database: str = None,
+        database: str,
         name_contains: str = None,
-        search_text: str = None,
     ) -> List[str]:
         """
-        Returns a list of tables in a Glue database.
+        Returns a list of tables in Amazon Redshift Spectrum database.
```
Trymzet, line 66: It was correct originally -- this is a Glue database, as per `awswrangler` docs. Spectrum reads data from Glue catalog.

Trymzet (changes requested):
Minor comment, rest is very good

viadot/sources/redshift_spectrum.py
AnnaGerlich: Indirectly it's also Amazon Redshift so I believe we could call it `Amazon Redshift Spectrum` so that it's more transparent for users. They're usually unaware of using Glue. WDYT?

Trymzet: Hm in fact both are incorrect, as `database` is actually any external database that user wants to query via Spectrum (it can be Athena, Hive, or Glue, or multiple instances of each etc.). For example, you could have `hive1` and `hive2` databases in Spectrum, which are both Hive databases. 

How about "Returns a list of tables in a specified external database of Amazon Redshift Spectrum."?

Trymzet (approved):
Looks good

PR #679, opened 2023-04-28
Dev add url customer gauge

hhnks:
## Summary
Added `url` argument to `customer_gauge` source class and added `enpoint_url` argument to task and flow classes. 

## Importance
User can pass his custom URL if it will change in future or will be created new endpoint.

## Checklist

This PR:

### DIFF
--- a/viadot/sources/customer_gauge.py
+++ b/viadot/sources/customer_gauge.py
@@ -44,13 +44,12 @@ class CustomerGauge(Source):
                 raise ValueError(
                     "Incorrect endpoint name. Choose: 'responses' or 'non-responses'"
                 )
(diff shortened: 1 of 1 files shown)


viadot/sources/customer_gauge.py
```diff
@@ -43,9 +45,12 @@ def __init__(
                     "Incorrect endpoint name. Choose: 'responses' or 'non-responses'"
                 )
         else:
```
Rafalz13, line 47: use `elif url is not None:`
`else: raise ValeError` at the end

Rafalz13 (commented):
Small changes.

PR #688, opened 2023-05-12
🐛 Fixed dir path

angelika233:
## Summary
Fixing issue when there is no path specified 

## Importance
fixing a bug

## Checklist

This PR:


viadot/sources/base.py
```diff
@@ -156,10 +156,12 @@ def to_parquet(
...
-
-        if not os.path.isfile(path):
-            directory = os.path.dirname(path)
-            os.makedirs(directory, exist_ok=True)
+        try:
+            if not os.path.isfile(path):
+                directory = os.path.dirname(path)
+                os.makedirs(directory, exist_ok=True)
+        except FileNotFoundError:
+            pass
```
Rafalz13, line 164: Just a small suggestion. What do you think about adding a logger there with the information that the file is not found?  To catch the exception and provide some information.

tests/unit/test_base.py
```diff
@@ -58,6 +58,13 @@ def test_to_excel():
     os.remove("testbase.xlsx")
 
 
+def test_to_parquet():
```
Rafalz13, line 61: Are you checking here for a 'file not found'? Please add a test to your `try except` block.

Rafalz13 (commented):
Looks good. Added suggestions.

PR #710, opened 2023-06-21
Outlook bug

Diego-H-S:
## Summary
- In the `subject` parameter occasionally appears an extra separator, which has been solved by replacing it with a space.
- When the length of the `receivers` parameter is longer than 8000 (maximum length for creating a `varchar`) the string is truncated to the latest email added to this `receivers` parameter.

## Importance
It happened in the past but I would be necessary to avoid similar situations in the future.

## Checklist

This PR:
- [x] links relevant issue(s)


viadot/sources/outlook.py
```diff
@@ -182,10 +182,13 @@ def _get_messages_from_mailbox(
                     recivers_list = fetched.get("toRecipients")
                     recivers = " "
                     if recivers_list is not None:
-                        recivers = ", ".join(
-                            reciver["emailAddress"]["address"]
-                            for reciver in recivers_list
-                        )
+                        for reciver in recivers_list:
+                            add_string = f", {reciver['emailAddress']['address']}"
+                            if sum(list(map(len, [recivers, add_string]))) >= 8000:
```
Rafalz13, line 187: Why do you assign `8000`?

viadot/sources/outlook.py
```diff
@@ -182,10 +182,13 @@ def _get_messages_from_mailbox(
...
+                                break
+                            else:
+                                recivers += add_string
```
Rafalz13, line 190: My suggestion - what do you think about using only 1 if to this case?
```suggestion
                            if sum(list(map(len, [recivers, add_string]))) < 8000:
                               recivers += add_string
                                
```

viadot/sources/outlook.py
```diff
@@ -182,10 +182,13 @@ def _get_messages_from_mailbox(
                     recivers_list = fetched.get("toRecipients")
                     recivers = " "
```
Rafalz13, line 183: Do the `receivers` need to have an empty space or can it be like `receivers=""`?

Rafalz13 (commented):
* Added comments.
* Please add tests for this functionality with `8000`.

viadot/sources/outlook.py
Diego-H-S: 8000 is the maximum to no overload data in a `varchar()` variable. More than that data is right truncated and the entire row is useless.

viadot/sources/outlook.py
Diego-H-S: The en is the same, because, I need to break the `for` loop when it reaches 8000, after that limit you are iterating without any purpose.
If I use your sentence, then I would need an `else: break` which is basically the same as it already there.

viadot/sources/outlook.py
Diego-H-S: I give an empty space to avoid any problem if finally, the program doesn't have any data to append.

Rafalz13: If you are sure about this option please add tests for that

Rafalz13 (commented):
Please add tests for new features:
* break if `>= 8000`
* handling `receivers`

viadot/sources/outlook.py
Diego-H-S: done!

Diego-H-S (commented):
created a new parameter and updated test file.

Rafalz13:
Please reduce logs to only directories where we can find some messages. In the previous version, a user was able to see everything. (It was a long log message)

Diego-H-S:
> Please reduce logs to only directories where we can find some messages. In the previous version, a user was able to see everything. (It was a long log message)

I have updated the source file to print out only folders with some messages.

PR #730, opened 2023-07-26
Add `VidClubToADLS` flow

hhnks:
## Summary
Flow class for ingesting data from Vid Club API to Azure Data Lake as .parquet or .csv file.

## Importance

With flow class we'll be able to schedule automated data downloads directly to ADLS.

## Checklist

This PR:
- [ ] links relevant issue(s)

### DIFF
--- a/viadot/flows/vid_club_to_adls.py
+++ b/viadot/flows/vid_club_to_adls.py
@@ -1,0 +1,196 @@
+import os
+from pathlib import Path
+from typing import Any, Dict, List, Literal
+import pandas as pd
+
+import pendulum
+from prefect import Flow
+from prefect.backend import set_key_value
+from prefect.utilities import logging
+
+from viadot.task_utils import (
+    add_ingestion_metadata_task,
+    df_get_data_types_task,
+    df_map_mixed_dtypes_for_parquet,
+    df_to_csv,
+    df_to_parquet,
+    dtypes_to_json_task,
+    update_dtypes_dict,
+)
+from viadot.tasks import AzureDataLakeUpload, VidClubToDF
+
+logger = logging.get_logger(__name__)
+
+
+class VidClubToADLS(Flow):
+    def __init__(
+        self,
+        name: str,
+        source: Literal["jobs", "product", "company", "survey"] = None,
+        from_date: str = "2022-03-22",
+        to_date: str = None,
+        items_per_page: int = 100,
+        region: Literal["bg", "hu", "hr", "pl", "ro", "si", "all"] = "all",
+        days_interval: int = 30,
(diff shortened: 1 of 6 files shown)


viadot/flows/vid_club_to_adls.py
```diff
@@ -0,0 +1,196 @@
...
+logger = logging.get_logger(__name__)
+
+
+class VidClubToADLS(Flow):
+    def __init__(
+        self,
+        name: str,
+        source: Literal["jobs", "product", "company", "survey"] = None,
+        from_date: str = "2022-03-22",
+        to_date: str = None,
```
Rafalz13, line 31: As I remember `to_date` was assigned in task as `today()`. Is it correct

viadot/flows/vid_club_to_adls.py
```diff
@@ -0,0 +1,196 @@
...
+        Then upload it to Azure Data Lake.
+
+        Args:
+            name (str): The name of the flow.
+            source (Literal["jobs", "product", "company", "survey"], optional): The endpoint source to be accessed. Defaults to None.
+            from_date (str, optional): Start date for the query, by default is the oldest date in the data 2022-03-22.
+            to_date (str, optional): End date for the query. Defaults to None.
+            items_per_page (int, optional): Number of entries per page. Defaults to 100.
+            region (Literal["bg", "hu", "hr", "pl", "ro", "si", "all"], optional): Region filter for the query. Defaults to "all". [July 2023 status: parameter works only for 'all' on API]
+            days_interval (int, optional): Days specified in date range per api call (test showed that 30-40 is optimal for performance). Defaults to 30.
```
Rafalz13, line 62: `api` -> `API`

Rafalz13 (commented):
Looks good

viadot/flows/vid_club_to_adls.py
hhnks: Actually only in task's init, it's `to_date: str = datetime.today()`, which doesn't make sense, because it's not string and in source there is already logic for `to_date == None`. 

So I can keep this as it is, and `to_date` by assigned as `datetime.today()` while executing source class or set default as `datetime.today().strftime("%Y-%m-%d")` and update defaults in task and source class.

Rafalz13: The flow will overwrite the task and task will overwrite the source parameter. If the flow has `None` the source will be the same.

hhnks: `to_date` parameter updated to have default value `None` and added comment in docstrings in flow, task and source classes:
"End date for the query. By default None, which will be executed as datetime.today().strftime("%Y-%m-%d") in code."

PR #744, opened 2023-09-12
Mindful auth

Diego-H-S:
## Summary
Mindful API has undergone a change in its structure. Since three weeks ago we are not able to access the API by adding credentials in the header. The solution given by the HCL support team is using `auth` parameter in the `request` function, and it was implemented in the present PR.

On the other hand, Instead of `Bearer` authentication, a new `Basic` authentication is used. The client chose this new way of authentication to avoid future problems with archived credentials after 90 days.

## Importance
Required by client. Very important.

## Checklist

This PR:
- [x] links relevant issue(s)

### DIFF
--- a/viadot/sources/mindful.py
+++ b/viadot/sources/mindful.py
@@ -1,11 +1,12 @@
 import os
-from datetime import datetime, timedelta
 from io import StringIO
-from typing import Any, Dict, Literal
+from datetime import datetime, timedelta
+from typing import Any, Dict, Literal, Tuple
 
 import pandas as pd
 import prefect
 from requests.models import Response
+from requests.auth import HTTPBasicAuth
 
 from viadot.exceptions import APIError
 from viadot.sources.base import Source
@@ -15,7 +16,7 @@ from viadot.utils import handle_api_response
 class Mindful(Source):
     def __init__(
         self,
-        header: str,
+        auth: Tuple[str],
         region: Literal["us1", "us2", "us3", "ca1", "eu1", "au1"] = "eu1",
         start_date: datetime = None,
         end_date: datetime = None,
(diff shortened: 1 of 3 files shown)


viadot/tasks/mindful.py
```diff
@@ -117,12 +117,13 @@ def run(
                 credentials_mindful = None
                 raise CredentialError("Credentials not found.")
 
-        header = {
-            "Authorization": f"Bearer {credentials_mindful.get('VAULT')}",
-        }
+        auth = (
+            credentials_mindful["CUSTOMER_UUID"],
+            credentials_mindful["AUTH_TOKEN"],
```
Rafalz13, line 122: Here you are getting `CUSTOMER_UUID` and `AUTH_TOKEN` but docstring in the source says that `The structure is user and password`. What is correct?

viadot/sources/mindful.py
```diff
@@ -27,7 +28,7 @@ def __init__(
         """Mindful connector which allows listing and downloading into Data Frame or specified format output.
 
         Args:
-            header (str): Header with credentials for calling Mindful API.
+            auth (Tuple[str]): Authentication credentials for calling Mindful API. The structure is user and password.
```
Rafalz13, line 31: The docstring says that `The structure is user and password`. In the task, you are getting `CUSTOMER_UUID` and `AUTH_TOKEN`. It doesn't look like a username and password.

Rafalz13 (commented):
Please modify tests for Mindful source and task if needed.

viadot/tasks/mindful.py
Diego-H-S: In mindful, the user is defined by an ID, in this case, `CUSTOMER_UUID`, and the password is defined by, what they said `AUTH TOKEN`, but this is a password, I  mean, if you need to generate a token for OAuth, this not would be correct, you would have to create a token.

Diego-H-S: when I saved it to the credentials, I added the same words they decided to use for the user (mindful velux user) and the password (mindful velux password).

viadot/sources/mindful.py
Diego-H-S: But the auth accepts the function `HTTPBasicAuth` which only needs the user and password.
I  you think it necessary for name consistency, I could replace `user` and `password` in docstring by `customer_uuid` and `auth_token` respectively.

Rafalz13 (approved):

PR #822, opened 2023-11-24
Fix url in vidclub connector

KaterynaIurieva:
Url for api endpoint  was changed, because parameters for endpoints got changed and value 'all' for parameter 'region' is no longer valid. Instead by default parameter 'region' (for job, company and survey) isn't used at all.

### DIFF
--- a/viadot/sources/vid_club.py
+++ b/viadot/sources/vid_club.py
@@ -263,7 +263,7 @@ class VidClub(Source):
         from_date: str = "2022-03-22",
         to_date: str = None,
         items_per_page: int = 100,
-        region: Literal["bg", "hu", "hr", "pl", "ro", "si", "all"] = "all",
+        region: Literal["bg", "hu", "hr", "pl", "ro", "si", "all"] = None,
         days_interval: int = 30,
     ) -> pd.DataFrame:
         """
--- a/viadot/sources/vid_club.py
+++ b/viadot/sources/vid_club.py
@@ -58,7 +58,7 @@ class VidClub(Source):
             api_url (str): Generic part of the URL to Vid Club API.
             items_per_page (int): number of entries per page.
             source (Literal["jobs", "product", "company", "survey"], optional): The endpoint source to be accessed. Defaults to None.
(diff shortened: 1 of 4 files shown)


viadot/sources/vid_club.py
```diff
@@ -207,14 +225,26 @@ def get_response(
...
+            response, first_url = self.check_connection(
+                source=source,
+                from_date=from_date,
+                to_date=to_date,
+                items_per_page=items_per_page,
+                region=region,
+            )
+        else:
+            raise ValidationError(
+                "Pick one these sources: jobs, product, company, survey"
```
adrian-wojcik, line 246:
```suggestion
                "Pick one of these sources: jobs, product, company, survey"
```

adrian-wojcik (changes requested):
Looks good.

viadot/sources/vid_club.py
```diff
@@ -207,14 +225,26 @@ def get_response(
             )
         if to_date == None:
             to_date = datetime.today().strftime("%Y-%m-%d")
+        if source in ["jobs", "product", "company"]:
+            response, first_url = self.check_connection(
+                source=source,
+                from_date=from_date,
+                to_date=to_date,
+                items_per_page=items_per_page,
```
Rafalz13, line 233: Why region is missing here? It can be a country code or None. If it is None it shouldn't be added to 'url' but I think it can be added here - it means that second if is not necessary here.

We already handling this case in the `build_query` function:
```if source in ["jobs", "product", "company"]: lines 69-72
            url = f"{api_url}{source}?from={from_date}&to={to_date}&region={region}&limit={items_per_page}"
        elif source == "survey":
            url = f"{api_url}{source}?language=en&type=question"```

**Review by Rafalz13 — COMMENTED** [review]
Look good. Please check if tests are missing or need to be updated.

**Review by Rafalz13 — COMMENTED** [review]
Please check tests and update if possible after adding this new feature.

File: viadot/tasks/vid_club.py [hunk 3]
```diff
@@ -85,7 +85,7 @@ def run(
         from_date: str = "2022-03-22",
         to_date: str = None,
         items_per_page: int = 100,
-        region: str = "all",
+        region: str = None,
```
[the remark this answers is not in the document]
KaterynaIurieva: done, and also changed default to none in total_load method for source

PR #835, opened 2023-12-19
✨ Added `SAPRFCV2` source class

AnnaGerlich:


m-paz (approved):

viadot/sources/sap_rfc.py
```diff
@@ -519,3 +651,513 @@ def to_df(self):
...
+                        row_index, data_raw, chunk, fields
+                    )
+                else:
+                    start = False
+
+                data_raw = catch_extra_separators(
+                    data_raw, record_key, sep, fields, self.replacement
+                )
+
+                records = np.array([row[record_key].split(sep) for row in data_raw])
```
trymzet, line 1128: Seems like this could be refactor to iterate through the data only once and apply `catch_extra_separators()` row-wise (no need to use `numpy` then as well).

trymzet (approved):

trymzet:
OK from me, other than the comment above I'd say this could be cleaned up a lot by having `SAPv2` inherit from `SAP` rather than source and by changing the SAPV2-specific functions into `SAPV2` methods. Then it's clear what V2 does differently.

PR #869, opened 2024-03-19
✨ Added `validate()` util for dataframe validation

burzekj:
## Summary
Logic for pd.DataFrame's validation due to bussines/dataset owner requirements on extraction level.
Added  @add_viadot_metadata_columns into saprfc connector.

## Checklist

This PR:
- [ ] links relevant issue(s)


viadot/utils.py
```diff
@@ -483,4 +484,153 @@ def wrapper(*args, **kwargs) -> pd.DataFrame:
 
 
 def get_fqn(table_name: str, schema_name: str = None) -> str:
-    return f"{schema_name}.{table_name}" if schema_name else table_name
\ No newline at end of file
+    return f"{schema_name}.{table_name}" if schema_name else table_name
+
+
+def validations(
```
trymzet, line 490: Function names should be verbs -- so here, it should be `validate()`.

viadot/utils.py
```diff
@@ -483,4 +484,153 @@ def wrapper(*args, **kwargs) -> pd.DataFrame:
...
+    Task to validate the data on DataFrame level. All numbers in the ranges are inclusive.
+    tests:
+        - `column_size`: dict{column: size}
+        - `column_unique_values`: list[columns]
+        - `column_list_to_match`: list[columns]
+        - `dataset_row_count`: dict: {'min': number, 'max', number}
+        - `column_match_regex`: dict: {column: 'regex'}
+        - `column_sum`: dict: {column: {'min': number, 'max': number}}
+    Args:
+        df (pd.DataFrame): The data frame for validation.
```
trymzet, line 506:
```suggestion
        df (pd.DataFrame): The dataframe to validate.
```

viadot/utils.py
```diff
@@ -483,4 +484,153 @@ def wrapper(*args, **kwargs) -> pd.DataFrame:
...
+
+    if logger is None:
+        logging.basicConfig(level=logging.INFO)
+        logger = logging.getLogger("prefect_shell.utils")
+
+    failed_tests = 0
+    failed_tests_list = []
+
+    if tests is not None:
+        if "column_size" in tests:
```
trymzet, line 520: I think each test should be split into a separate function and here you should just be calling the correct functions. This function is already too long and it will only grow. Also, the way this is done here, there's no way to unit test specific tests.

viadot/utils.py
```diff
@@ -483,4 +484,153 @@ def wrapper(*args, **kwargs) -> pd.DataFrame:
...
 
 def get_fqn(table_name: str, schema_name: str = None) -> str:
-    return f"{schema_name}.{table_name}" if schema_name else table_name
\ No newline at end of file
+    return f"{schema_name}.{table_name}" if schema_name else table_name
+
+
+def validations(
+    df: pd.DataFrame,
+    tests: dict = None,
```
trymzet, line 492: Nit: it would be better to have either an Enum or a model for this, which would allow IDE to display suggestions and documentation. Eg.

```python
from pydantic import BaseModel

class DataFrameTests(BaseModel):
    """Available DataFrame tests."""
    column_size: dict[str, int] | None
    column_unique_values: list[str]
    ...

def validate(..., tests: DataFrameTests = None):
    ...
```

(just FYI, no need to add it now)

viadot/utils.py
```diff
@@ -483,4 +484,153 @@ def wrapper(*args, **kwargs) -> pd.DataFrame:
...
+                        level=stream_level,
+                        msg=f"[column_sum] Sum of {col_sum} for {column} is out of the expected range - <{min_bound}:{max_bound}>",
+                    )
+    else:
+        return "No dataframe tests to run."
+
+    if failed_tests > 0:
+        failed_tests_msg = ", ".join(failed_tests_list)
+        raise ValidationError(
+            f"Validation failed for {failed_tests} test/tests: {failed_tests_msg}"
```
trymzet, line 635:
```suggestion
            f"Validation failed for {failed_tests} test(s): {failed_tests_msg}"
```

trymzet (changes requested):
Looks mostly good, main thing is to refactor the big func so that each test is a function, and also rename it to be a verb

PR #880, opened 2024-04-02
♻️ Added Databircks as optional dependency and source

djagoda881:
## Summary

Databricks source is now an optional source and dependence, which decreases the size of the viadot package and viadot docker image. This PR includes:
- Changes in the Databricks source.
- Changes in Databricks tests and SAP_RFC by which they are skipped if optional dependencies are not installed.
- Entry in `CONTRIBUTING.md` explaining the release process. 
- Changes in Dockerfile that skip the installation of Databricks packages if the build argument is not passed.
- Changes in CI build docker image.  
- Moved tests related to the source into the `tests/integration/` directory.

## Checklist

This PR:
- [ ] links relevant issue(s)


trymzet:
@djagoda881 pls rebase and fix the pre-commit errors. Most can be fixed automatically just by running `pre-commit run --all-files` (first you need to install pre-commit with `rye install pre-commit` and enable the hooks on the viadot repo by running `pre-commit install` inside the viadot repo's root dir. We probably need to add these pre-commit instructions to `CONTRIBUTING.md` BTW.

src/viadot/sources/__init__.py
```diff
@@ -26,3 +24,6 @@
 
 if find_spec("pyrfc"):
     __all__.extend(["SAPRFC", "SAPRFCV2"])
+
+if find_spec("databricks-connect"):
+    __all__.extend(["Databricks"])
```
trymzet, line 29: if it's a single item you can just use `append()`

docker/Dockerfile
```diff
@@ -53,14 +53,20 @@ RUN pip install --upgrade pip
...
     mkdir -p /etc/apt/keyrings && \
     wget -O - https://packages.adoptium.net/artifactory/api/gpg/key/public | tee /etc/apt/keyrings/adoptium.asc && \
     echo "deb [signed-by=/etc/apt/keyrings/adoptium.asc] https://packages.adoptium.net/artifactory/deb $(awk -F= '/^VERSION_CODENAME/{print$2}' /etc/os-release) main" | tee /etc/apt/sources.list.d/adoptium.list && \
-    apt update -q && \
-    apt install -q temurin-11-jdk -y && \
-    find /usr/bin/java -type d -exec chmod 777 {} \;
+    apt-get update && \
+    apt-get install -y temurin-11-jdk && \
+    find /usr/bin/java -type d -exec chmod 777 {} \; && \
+    pip install .[databricks]; \
```
trymzet, line 68: add `--user`

trymzet, line 68:
```suggestion
    pip install .[databricks] --user; \
```

src/viadot/utils.py
```diff
@@ -243,84 +241,6 @@ def _cast_df_cols(
     return df
 
 
-def build_merge_query(
```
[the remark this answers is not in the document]
djagoda881: I've checked salesforce source and it doesn't use `build_merge_query`. Btw this functions uses spark.dataframe and some internal spark fuctions

src/viadot/sources/__init__.py
djagoda881: [🎨 Changed list method in __init__.py](https://github.com/dyvenia/viadot/pull/880/commits/53597e0654b95df4b3e5f5035e42df626546af5a)

src/viadot/utils.py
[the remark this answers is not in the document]
trymzet: It's used by an `AzureSQL` task https://github.com/dyvenia/viadot/blob/8d12bc165399e4f32e53cd21fc044ab6a42201ea/viadot/tasks/azure_sql.py#L14
which is used by some production flows in Prefect 1.

You can find it by typing "build_merge_query" into the search bar

Remember about Chesterton's fence -- if you don't understand something, don't touch it

src/viadot/sources/__init__.py
trymzet: Yes but you need to append an item, not a list... `a.append("b")`, not `a.append(["b"])`. That's the difference between `append()` and `extend()`.

trymzet (changes requested):
Waiting for existing comments to be addressed, once done I'll re-review

src/viadot/sources/__init__.py
djagoda881: [🐛 Fixed imports in __init__.py](https://github.com/dyvenia/viadot/pull/880/commits/de5124f7cad07eb58db8773de4dad7ca9df8d665)

djagoda881: I added imports in the if statement because, in the case of installed, e.g. `pyspark` it was added Databricks to `__all__ `variable but it still didn't imported a Databricks class, which caused an error. So, for example, if we installed Databricks, we couldn't use it.

src/viadot/utils.py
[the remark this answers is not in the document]
djagoda881: It's used but in viadot1. I was looking in viadot2, but yeah it's good idea to keep it in utils.py
[♻️ Moved build_merge_query() to utils.py](https://github.com/dyvenia/viadot/pull/880/commits/7cdcf26a2389eade20f3e702d38a63046a8b4bf9)

docker/Dockerfile
```diff
@@ -45,14 +45,20 @@ COPY docker/odbcinst.ini /etc
 # so `pip install pyrfc` breaks if all deps are not already present.
 RUN pip install cython==0.29.24 --user
 
+ARG INSTALL_DATABRICKS=false
+
 # BEGIN Databricks source setup
-RUN apt install -y wget apt-transport-https && \
+RUN if [ "$databricks" = "true" ]; then \
```
trymzet, line 51: shouldn't this be changed to if INSTALL_DATABRICKS?

src/viadot/utils.py
```diff
@@ -18,7 +17,95 @@
...
+
+try:
+    import pyspark.sql.dataframe as spark
+
+    from viadot.sources import Databricks
+except ImportError:
+    pass
+
+
+def _get_table_columns(
```
trymzet, line 29: I don't think this file should be changed at all; the solution to the Databricks import is [already implemented](https://github.com/dyvenia/viadot/blob/2.0/src/viadot/utils.py) in `2.0` branch (I added it in the mean time since I needed it for something else).

trymzet (changes requested):
Added comments

src/viadot/utils.py
djagoda881: [♻️ Refactored databricks related functions](https://github.com/dyvenia/viadot/pull/880/commits/c222d003180c954aa3cda6eed22c007fbd0c78a8)
Changed only spark import because its raise a exception when databricks is not installed

docker/Dockerfile
djagoda881: [🐛 Fixed bug with inaccessible packages](https://github.com/dyvenia/viadot/pull/880/commits/f4469106af842a4fd1890e0430639a1aa827dcfc)

trymzet: 1. You inserted this block in the middle of viadot set up, can you move it above or below it?
https://github.com/dyvenia/viadot/blob/f4469106af842a4fd1890e0430639a1aa827dcfc/docker/Dockerfile#L59-L73

3. Just use if/else here

https://github.com/dyvenia/viadot/blob/f4469106af842a4fd1890e0430639a1aa827dcfc/docker/Dockerfile#L76-L79

djagoda881: [🎨 Improved Dockerfile format](https://github.com/dyvenia/viadot/pull/880/commits/85145a7d614d915f20bd60ec188cc241537fc9db)

trymzet (approved):

PR #899, opened 2024-05-07
🚀 Added new functionality to saprfc source regarding where statement

adrian-wojcik:
## Summary

Adding new functionality to saprfc source regarding where statement and dynamic part of after 70 characters 

## Importance

It is important to prevent passing sql query without white spaces between operators and `WHERE` statement and also raise error when dynamic query is detected after 70 characters after `WHERE`

## Checklist

This PR:
- [X ] follows the guidelines laid out in `CONTRIBUTING.md`
- [ ] links relevant issue(s)
- [X ] adds/updates docstrings (if appropriate)
- [ X] adds an entry in `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `Trino` source
 - Added `MinIO` source
 - Added `gen_split()` method to `SAPRFCV2` class to allow looping over a data frame with generator - improves performance
+- Added `adjust_where_condition_by_adding_missing_spaces()` to `SAPRFC`. The function that is checking raw sql query and modifing it - if needed.
 
 ### Changed
 
--- a/src/viadot/sources/sap_rfc.py
+++ b/src/viadot/sources/sap_rfc.py
@@ -33,6 +33,35 @@ from viadot.utils import add_viadot_metadata_columns, validate
 logger = logging.getLogger()
 
 
+def make_where_statement_fine(sql: str) -> str:
+
+    # Check if 'WHERE' statement is not attached to 'FROM' or column name as there is need for space " " on both side of 'WHERE'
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `Trino` source
 - Added `MinIO` source
 - Added `gen_split()` method to `SAPRFCV2` class to allow looping over a data frame with generator - improves performance
+- Added `adjust_where_condition_by_adding_missing_spaces()` to `SAPRFC`. The function that is checking raw sql query and modifing it - if needed.
```
marcinpurtak, line 31: Name of this function doesn't appear in the code - I can see 
 `make_where_statement_fine`  instead

marcinpurtak (changes requested):

CHANGELOG.md
adrian-wojcik: Changed

marcinpurtak (approved):

trymzet (commented):
I don't have anything to say about the code itself, but why isn't this logic added to `_get_where_condition()`?

adrian-wojcik:
> I don't have anything to say about the code itself, but why isn't this logic added to `_get_where_condition()`?

Because, in method `query()` variable sql is passed to multiple methods, so if we will add this functionality to the `_get_where_condition()` other functions will fail. So we decide that better approach is to take raw sql string at the beginning and refactor it before passing to other methods

trymzet:
Ah ok, I thought this is only sanitizing the `WHERE` clause, didn't think this would impact the rest of the parsing (I assume you mean that the rest of the parsing in `extract_values()` is failing, right?). 

In this case, I would only recommend adding a longer docstring with some context as to why this is needed.

trymzet (approved):
Looks good. 

@adrian-wojcik just a tip: you can use tools like Grammarly or gen AI chatbots like ChatGPT to fix the English grammar in docstrings/comments.

PR #914, opened 2024-05-31
Modified reading the file and inspect provided URL

Rafalz13:
## Summary
- Modified file reading to only read `XLSX files`.
- Added functionality to read and return the endpoint type in a URL, depending on whether it is a file or directory.
- Added support for `NA` values ​​from Excel to be treated as a string value, rather than `Nan` as was previously the case.
- Mapped all columns to string type to avoid problems that previously existed with treating `str` as an `int`.

## Importance
It is needed for the prefect_viadot sharepoint task.

## Checklist

This PR:
- [ ] links relevant issue(s)


trymzet (approved):

trymzet:
@Rafalz13 the `test_connection_with_genesys_api()` test shouldn't be in unit tests since it requires connection to Genesys API, can you remove it pls so that the tests CI passes?

marcinpurtak (approved):

PR #926, opened 2024-06-13
🚀 [migration viadot-2-new-structure-migration] Add Eurostat source wi…

adrian-wojcik:
…th tests

## Summary
Adding modified Eurostat source for Prefect 2.0 with tests

## Importance
Because, it is adding another source to be able to perform on Prefect 2.0

## Checklist

This PR:
- [X ] follows the guidelines laid out in `CONTRIBUTING.md`
- [x] links relevant issue(s)
- [X ] adds/updates tests (if appropriate)
- [X ] adds/updates docstrings (if appropriate)
- [X ] adds an entry in `CHANGELOG.md`


src/viadot/orchestration/prefect/tasks/eurostat.py
```diff
@@ -0,0 +1,46 @@
...
+
+from prefect import task
+
+
+@task(retries=3, retry_delay_seconds=10, timeout_seconds=60 * 60)
+def eurostat_to_df(
+    dataset_code: str,
+    params: dict = None,
+    base_url: str = "https://ec.europa.eu/eurostat/api/dissemination/statistics/1.0/data/",
+    requested_columns: list = None,
```
trymzet, line 11: nit: maybe just `columns`?

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+        params_and_codes = {}
+        for key in available_params:
+            if key in dimension:
+                codes = list(dimension[key]["category"]["index"].keys())
+                params_and_codes[key] = codes
+        return params_and_codes
+
+    def make_params_validation(self):
+        """Function for validation of given parameters in comparison
+        to parameteres and their codes from JSON.
```
trymzet, line 66: 1. Docstrings should be in imperative (this applies to all the other docstrings as well but I'm not going to repeat this comment everywhere to not pollute)
2. Sorry, I didn't understand anything more than "this validates parameters". I'd say it's best to either be brief (like below suggestion) or provide a summary of your validation logic (list things you're validating).

```suggestion
        """Validate API request parameters."""
```

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+        self,
+        dataset_code: str,
+        params: dict = None,
+        base_url: str = "https://ec.europa.eu/eurostat/api/dissemination/statistics/1.0/data/",
+        requested_columns: list = None,
+        if_empty: str = "fail",
+        tests: dict = None,
+    ) -> pd.DataFrame:
+        """Function responsible for getting response, creating DataFrame using method 'eurostat_dictionary_to_df'
+           with validation of provided parameters and their codes if needed.
```
trymzet, line 194: Sry just noticed it here but -- by "codes", you mean "currency codes", right? If so, can you use that full form in the documentation (and consider also using it in variable names)? It would make it easier to understand.

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+                - `dataset_row_count`: dict: {'min': number, 'max', number}
+                - `column_match_regex`: dict: {column: 'regex'}
+                - `column_sum`: dict: {column: {'min': number, 'max': number}}
+        Raises:
+            APIError: If there is an error with the API request.
+            ValueError: If the resulting DataFrame is empty.
+            TypeError: If self.params is different type than a dictionary.
+            TypeError: If self.requested_columns is different type than a list containing strings.
+        Returns:
+            pd.DataFrame: Final DataFrame or raise prefect.logger.error, if issues occur.
```
trymzet, line 222: Remove mention of prefect, also I don't think this is accurate?

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+            self.make_params_validation()
+
+        if data_frame.empty:
+            try:
+                self.make_params_validation()
+                self._handle_if_empty(if_empty)
+            except SKIP:
+                return pd.DataFrame()
+
+        # merging data_frame with label and last updated date
```
trymzet, line 250: This is not a strict requirement, but I would say, pick one style and stick to it throughout your code (so eg. either always uppercase or never).

IMHO it's best if the comments are written like sentences, with standard English grammar and in imperative mood (same as docstrings):

```suggestion
        # Merge data_frame with label and last updated date.
```

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+        try:
+            response = handle_api_response(self.base_url, params=self.params)
+            data = response.json()
+            data_frame = self.eurostat_dictionary_to_df(["geo", "time"], data)
+        except APIError:
+            self.make_params_validation()
+
+        if data_frame.empty:
+            try:
+                self.make_params_validation()
```
trymzet, line 245: Why not call this at the very beginning of `to_df()`? You don't need to call the API to know whether the params are correct or not (in fact thanks to this validation func, you know in advance if an API call will fail and so you can fail faster).

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+                    f"Available columns: {' | '.join(columns_list)}"
+                )
+                raise ValueError("Provided columns are not available!")
+
+            data_frame = data_frame.loc[:, needed_column_after_validation]
+
+            # Additional validation from utils
+            validate(df=data_frame, tests=tests)
+
+            return data_frame
```
trymzet, line 290: 1) Would be good to put this code in some helper method because this `to_df()` is already pretty long (~110 lines; usually it's best to stay within ~50 (you should see the entire function without scrolling)).

2) Why do you only run `validate()` if `self.requested_columns` is specified, shouldn't it be executed in either case?

tests/viadot_tests/integration/test_eurostat.py
```diff
@@ -0,0 +1,226 @@
...
+import pytest
+from viadot.sources import Eurostat
+
+
+def test_and_validate_dataset_code_without_params(caplog):
+    """This function is designed to test the accuracy of the data retrieval feature in a program.
+    Specifically, it tests to ensure that the program returns a non-empty DataFrame when a correct
+    dataset code is provided without any parameters. The function is intended to be used in software
+    development to verify that the program is correctly retrieving data from the appropriate dataset.
+    """
```
trymzet, line 13: So basically

```suggestion
    """Ensure that a non-empty DataFrame is returned when a correct dataset code is provided
    without any parameters
    """
```

These test docstrings are way too verbose, can't ChatGPT it sum it up in one-two lines? 😄

tests/viadot_tests/integration/test_eurostat.py
```diff
@@ -0,0 +1,226 @@
...
+    task = Eurostat()
+
+    with pytest.raises(ValueError, match="DataFrame is empty!"):
+        with caplog.at_level(logging.ERROR):
+            task.to_df(dataset_code="ILC_DI04E")
+    assert (
+        "Failed to fetch data for ILC_DI04E, please check correctness of dataset code!"
+        in caplog.text
+    )
```
trymzet, line 36: In general it's cleaner to validate exceptions rather than logging output, I would treat validating logs as a last resort (it's used to validate that the logging is correct, not that the code's logic is correct).

trymzet (changes requested):
First of all, this is a good PR. Since I don't think I've had the chance to review your code before, I'll be a bit more detailed, so you can pick up a few things for your future contributions.

Other than specific comments in the code, below are some general remarks:

1. It's a bad practice to store variables as instance attributes without an explicit need to do so. This makes the code harder to follow and test. Also, it should be very rare to assign any attributes outside of the `__init__()` function. You can see it for example in your `make_params_validation()` method -- it's a lot easier to test if it accepts a `parameters` param rather than using the attribute. Because instead of this:

```python
def test_make_params_validation():
   test_source = Eurostat()
   test_failing_params = {...}
   with pytest.raises(ValueError):
       test_source.to_df(params=test_failing_params, ...)
```
(So basically this way you can't even test the method directly, because you set `self.params` inside `to_df()` method. And since you need to use `to_df()` to test anything, every test becomes an integration test.)

Instead, it's easier to have this:

```python
def test_make_params_validation():
    test_source = Eurostat()
    test_failing_params = {...}
    with pytest.raises(ValueError):
        test_source.make_params_validation(parameters=test_failing_params)
```

As you can see, restraining from storing non-instance related variables in `self` allows you to think about and use methods in isolation. One way to think about it is "is this thinnk related to the object or an action on the object", and eg. in this case, it's clear that `params` are not related to the Eurostat API in general, but to a specific call to the Eurostat API.

2. Please remember about adding unit tests to sources. Unit tests are executed automatically on each PR to the `2.0` branch, ensuring that the code added in the PR works, but also that it does not break existing code. So it's important to start adding as many such tests as possible to enable frequent releases without production issues. In this case, I think you should unit test at least the  `make_params_validation()` and `eurostat_dictionary_to_df()` methods, but looking at your test suite in the integration tests, you might be able to convert more of them into unit tests.

3. I don't like the directory structure of the `tests/` directory -- I think we should keep the  unit/integration separation. Can you change the structure to below?
```console
tests/unit
tests/integration
```

Since orchestration tests are all integration tests, we can put it inside a `tests/integration/orchestration` subfolder.

4. FYI we use pre-commit (see our [CONTRIBUTING](https://github.com/dyvenia/viadot/blob/2.0-new-repository-structure/CONTRIBUTING.md) guide) to ensure some standard formatting. It runs in a CI job on each PR to 2.0 branch, but it's also a good idea to install it locally as described in the guide.

trymzet:
@adrian-wojcik @fdelgadodyvenia also, question: do we still need the Databricks stuff?

src/viadot/orchestration/prefect/tasks/eurostat.py
adrian-wojcik: Changed to `columns`.

src/viadot/orchestration/prefect/tasks/eurostat.py
```diff
@@ -0,0 +1,46 @@
...
+        tests:
+            - `column_size`: dict{column: size}
+            - `column_unique_values`: list[columns]
+            - `column_list_to_match`: list[columns]
+            - `dataset_row_count`: dict: {'min': number, 'max', number}
+            - `column_match_regex`: dict: {column: 'regex'}
+            - `column_sum`: dict: {column: {'min': number, 'max': number}}
+
+    Returns:
+        pd.DataFrame: None empty DataFrame.
```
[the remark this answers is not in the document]
adrian-wojcik: You're right. Changed to `Pandas DataFrame`

src/viadot/sources/eurostat.py
adrian-wojcik: Changed.

src/viadot/sources/eurostat.py
adrian-wojcik: Remmoved

src/viadot/sources/eurostat.py
adrian-wojcik: No, codes reffer to `codes` of specific parmeter e.g (for dataset "prc_hicp_midx" we could filetr dataset by specific item. "Food and non-alcoholic beverages" has code "CP01" - and this code is needed for filtering the data) in dataset that could be used to retriev specific values from dataset. It can be currency code as well, but not only.

src/viadot/sources/eurostat.py
adrian-wojcik: Changed.

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # Assigning list of available codes to specific parameters
+        params_and_codes = {}
+        for key in available_params:
+            if key in dimension:
+                codes = list(dimension[key]["category"]["index"].keys())
+                params_and_codes[key] = codes
+        return params_and_codes
+
+    def make_params_validation(self):
```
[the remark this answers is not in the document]
adrian-wojcik: Not really. This function is not doing validation itself. It's actually returning dictionary with available parameters codes for specific dataset, that are used in function make_params_validation() for validation. So I belive it is okey as it is.

src/viadot/sources/eurostat.py
adrian-wojcik: 1. I put requested_columns logic into method: requested_columns_validation().
2. I belive it is not, as validate() is outside if statement of requested_columns:
        # Validation and transformation of requested column
        if self.columns is not None:
            self.requested_columns_validation(data_frame=data_frame)

        # Additional validation from utils
        validate(df=data_frame, tests=self.tests)

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+            raise TypeError("Params should be a dictionary.")
+
+        self.base_url = f"{base_url}{self.dataset_code}?format=JSON&lang=EN"
+        self.requested_columns = requested_columns
+        if not isinstance(requested_columns, list) and requested_columns is not None:
+            raise TypeError("Requested columns should be provided as list of strings.")
+
+        # getting response from API
+        try:
+            response = handle_api_response(self.base_url, params=self.params)
```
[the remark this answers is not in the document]
adrian-wojcik: After, I moved all attributes to constructior - it can stay as it is. With what I read regarding OOP if I am using attribute of a class inside method I should reffer to it by self.

src/viadot/sources/eurostat.py
```diff
@@ -0,0 +1,290 @@
...
+        Returns:
+            pd.DataFrame: Final DataFrame or raise prefect.logger.error, if issues occur.
+        """
+
+        self.dataset_code = dataset_code
+        self.params = params
+        if not isinstance(self.params, dict) and self.params is not None:
+            raise TypeError("Params should be a dictionary.")
+
+        self.base_url = f"{base_url}{self.dataset_code}?format=JSON&lang=EN"
```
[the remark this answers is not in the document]
adrian-wojcik: Changed.

src/viadot/sources/eurostat.py
adrian-wojcik: True, this could be better approach.

tests/viadot_tests/integration/test_eurostat.py
adrian-wojcik: Done

tests/viadot_tests/integration/test_eurostat.py
adrian-wojcik: I am validating exceptions and logging at the same time

src/viadot/orchestration/prefect/tasks/eurostat.py
```diff
@@ -36,11 +35,11 @@ def eurostat_to_df(
...
     Returns:
-        pd.DataFrame: None empty DataFrame.
+        pd.DataFrame: Pandas DataFrame.
     """
 
-    data_frame = Eurostat().to_df(
-        dataset_code, params, base_url, requested_columns, tests
-    )
+    data_frame = Eurostat(
+        dataset_code=dataset_code, params=params, columns=columns, tests=tests
```
trymzet, line 42: These should be params to the `to_df()` method.

src/viadot/sources/eurostat.py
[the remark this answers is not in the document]
trymzet: This is about `make_params_validation()` - maybe you need to re-read my comment?

src/viadot/sources/eurostat.py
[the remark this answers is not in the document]
trymzet: Yes, but as I explained in another comment, storing things that should be parameters to your methods as class attributes is a bad practice, for reasons explained in that comment.

src/viadot/sources/eurostat.py
trymzet: Thanks. Can you pls also apply this rule to the rest of them, as per point 1.?

src/viadot/sources/eurostat.py
trymzet: Ah, okay. Is there a list of these codes you could link to in the docstring, or is this info only in your head? Or at least sth explaining what the codes are -- just like you explained to me above? Since this info is not possible to understand from the code itself.

src/viadot/sources/eurostat.py
trymzet: Hm actually I realized now you're actually calling the API every time to validate the params. But still, it seems like this belongs to `to_df()` as `params` should be a param of the `to_df()` method.

src/viadot/sources/eurostat.py
trymzet: AD 2) Sorry, this is unreadable. For the future, you can paste code samples using the code block Markdown syntax: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting

But I checked the code and see you have fixed the logic: you're now running `validate()` regardless of whether you're filtering any columns or not (here you were only validating if you were filtering out some columns).

trymzet (changes requested):
Addressed all comments.

One more ask - if you implemented sth that was requested in a comment, can you pls paste link to the commit where you implemented the change in your response (instead of only things like "Done")? It will make it much easier for me to check if/how something we discussed was implemented.

src/viadot/sources/eurostat.py
```diff
@@ -2,18 +2,24 @@
...
     """
 
+    base_url = "https://ec.europa.eu/eurostat/api/dissemination/statistics/1.0/data/"
+
     def __init__(
         self,
+        dataset_code: str,
+        params: dict = None,
+        columns: list = None,
+        tests: dict = None,
```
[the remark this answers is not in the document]
adrian-wojcik: https://github.com/dyvenia/viadot/pull/926/commits/6823c7cd8e033209d4f11abaf9d01c836a577158

src/viadot/sources/eurostat.py
```diff
@@ -183,71 +276,34 @@ class T_SIGNAL:
     @add_viadot_metadata_columns
     def to_df(
         self,
-        dataset_code: str,
-        params: dict = None,
-        base_url: str = "https://ec.europa.eu/eurostat/api/dissemination/statistics/1.0/data/",
-        requested_columns: list = None,
-        if_empty: str = "fail",
-        tests: dict = None,
```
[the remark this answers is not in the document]
adrian-wojcik: https://github.com/dyvenia/viadot/pull/926/commits/6823c7cd8e033209d4f11abaf9d01c836a577158

src/viadot/sources/eurostat.py
```diff
@@ -183,71 +276,34 @@ class T_SIGNAL:
...
+        if self.params is not None:
+            self.make_params_validation(
+                dataset_code=self.dataset_code, url=self.url, params=self.params
+            )
 
-        # getting response from API
+        # Getting response from API
         try:
-            response = handle_api_response(self.base_url, params=self.params)
+            response = handle_api_response(self.url, params=self.params)
```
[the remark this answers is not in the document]
adrian-wojcik: [[6823c7c](https://github.com/dyvenia/viadot/pull/926/commits/6823c7cd8e033209d4f11abaf9d01c836a577158)](https://github.com/dyvenia/viadot/pull/926/commits/6823c7cd8e033209d4f11abaf9d01c836a577158)

src/viadot/orchestration/prefect/tasks/eurostat.py
adrian-wojcik: https://github.com/dyvenia/viadot/pull/926/commits/cd1e3ec7731f4caa6ff0619a23c88f48c933e996

src/viadot/sources/eurostat.py
```diff
@@ -115,12 +167,53 @@ def make_params_validation(self):
...
+            self.logger.error(
+                f"Name of the columns: '{' | '.join(non_available_columns)}' are not in DataFrame. Please check spelling!\n"
+                f"Available columns: {' | '.join(data_frame.columns)}"
+            )
+            raise ValueError("Provided columns are not available!")
+
+        # Selecting only needed columns from the original DataFrame
+        data_frame = data_frame.loc[:, needed_columns]
+
+        return data_frame
```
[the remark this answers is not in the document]
adrian-wojcik: https://github.com/dyvenia/viadot/pull/926/commits/c7246c36a1c36898f0f72dc53c5cd9aef897253d

tests/unit/test_eurostat.py
```diff
@@ -0,0 +1,178 @@
+import logging
```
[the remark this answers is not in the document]
adrian-wojcik: https://github.com/dyvenia/viadot/pull/926/commits/ffdba7ced0d82f667f1c67d7f0fa81b6bcc183fc

trymzet:
This also seems to be created from an incorrect branch since it shows 111 modified files, while this should be more like 5?

PR #992, opened 2024-08-26
✨ Add `sql_server_to_parquet` flow

angelika233:
## Summary
Adeed sql_serer_to_parquet flow and fixed bcp task

## Importance
new connector and fix

## Checklist

This PR:
- [ ] links relevant issue(s)

### DIFF
--- a/src/viadot/orchestration/prefect/tasks/bcp.py
+++ b/src/viadot/orchestration/prefect/tasks/bcp.py
@@ -28,13 +28,13 @@ def bcp(
         path (str):  Where to store the CSV data dump used for bulk upload to a database.
         schema (str, optional): Destination schema. Defaults to None.
         table (str, optional): Destination table. Defaults to None.
-        chunksize (int, optional): Size of a chunck to use in the bcp function.
+        chunksize (int, optional): Size of a chunk to use in the bcp function.
             Defaults to 5000.
         error_log_file_path (string, optional): Full path of an error file. Defaults
             to "./log_file.log".
         on_error (str, optional): What to do in case of a bcp error. Defaults to "skip".
         credentials_secret (str, optional): The name of the secret storing
-            the credentialsto the SQLServer. Defaults to None.
+            the credentials to the SQLServer. Defaults to None.
             More info on: https://docs.prefect.io/concepts/blocks/
         credentials (dict[str, Any], optional): Credentials to the SQLServer.
             Defaults to None.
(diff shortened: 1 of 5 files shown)


src/viadot/orchestration/prefect/tasks/bcp.py
```diff
@@ -64,8 +70,33 @@ def bcp(
...
+        "-U",
+        uid,
+        "-P",
+        pwd,
+        "-b",
+        str(chunksize),
+        "-m",
+        str(max_error),
+        "-c",
+        "-v",
```
trymzet, line 93: So the only difference is the `-v` flag, which according to docs, shows the version of the tool? 😅 https://learn.microsoft.com/en-us/sql/tools/bcp-utility?view=sql-server-ver16&tabs=windows#-v

Also, why replace `shell_run_command()` with `subprocess.run()`?

src/viadot/orchestration/prefect/tasks/sql_server.py
```diff
@@ -109,7 +110,7 @@ def sql_server_query(
     credentials_secret: str | None = None,
     credentials: dict[str, Any] | None = None,
     config_key: str | None = None,
-) -> list[Record] | bool:
+) -> None:
```
trymzet, line 113: I think this was correct.

src/viadot/orchestration/prefect/tasks/bcp.py
```diff
@@ -1,34 +1,37 @@
...
-    table: str | None = None,
+    path: str,
+    schema: str,
+    table: str,
     chunksize: int = 5000,
     error_log_file_path: str = "./log_file.log",
     on_error: Literal["skip", "fail"] = "skip",
     credentials_secret: str | None = None,
     config_key: str | None = None,
+    credentials: dict[str, Any] | None = None,
```
trymzet, line 23: Let's not expose `credentials` in tasks/flows since Prefect stores parameter values in plain text and can also display them in logs, which is a security issue.

trymzet (changes requested):

src/viadot/orchestration/prefect/tasks/bcp.py
angelika233: I don't know why but `shell_run_command()` didn't work for this command

angelika233: Also in this case casting int parameters as str was needed

trymzet (approved):

PR #1002, opened 2024-08-26
2.0 epicor migration

angelika233:
## Summary
Migration of epicor connector

## Importance
migration

## Checklist

This PR:
- [ ] links relevant issue(s)


tests/integration/test_epicor.py
```diff
@@ -0,0 +1,53 @@
+import pandas as pd
```
trymzet, line 1: I feel like these tests aren't enough; key logic of this connector is not tested here. We should have a test for each of the 4 parsing functions. They should be easy to write, though - an LLM can probably autogenerate some test XML.

src/viadot/sources/epicor.py
```diff
@@ -0,0 +1,497 @@
...
+        Returns:
+            str: Output url string.
+        """
+        return (
+            "http://"
+            + self.credentials["host"]
+            + ":"
+            + str(self.credentials["port"])
+            + self.base_url
+        )
```
trymzet, line 451: nit: you can just compute this in the init and have it as an attribute, eg.
```python
self.url = "http://" + validated_creds["host"]+ ":" + str(validated_creds["port"]) + base_url
```

src/viadot/sources/epicor.py
```diff
@@ -0,0 +1,497 @@
...
+        """
+        raw_creds = credentials or get_source_credentials(config_key) or {}
+        validated_creds = EpicorCredentials(**raw_creds).dict(
+            by_alias=True
+        )  # validate the credentials
+
+        self.credentials = validated_creds
+        self.config_key = config_key
+        self.base_url = base_url
+        self.filters_xml = filters_xml
```
trymzet, line 408: I think this is per-request param, so should be a parameter in relevant methods (eg. `validate_filter()`, `get_xml_response()`)  rather than attribute.

It will also make the methods easier to unit test since they won't depend on state (eg. you could test `validate_filter()` by passing various test filters as a param, rather than creating separate `Epicor` instances for each test).

src/viadot/orchestration/prefect/tasks/epicor.py
```diff
@@ -0,0 +1,65 @@
...
+def epicor_to_df(
+    base_url: str,
+    filters_xml: str,
+    validate_date_filter: bool = True,
+    start_date_field: str = "BegInvoiceDate",
+    end_date_field: str = "EndInvoiceDate",
+    credentials_secret: str | None = None,
+    config_key: str | None = None,
+) -> pd.DataFrame:
+    """Load the result of a SQL Server Database query into a pandas DataFrame.
```
trymzet, line 23: 🤔

src/viadot/orchestration/prefect/flows/epicor_to_parquet.py
```diff
@@ -0,0 +1,64 @@
...
+    epicor_config_key: str | None = None,
+) -> None:
+    """Download a pandas `DataFrame` from Epicor Prelude API load it into Parquet file.
+
+    Args:
+        path (str): Path to Parquet file, where the data will be located.
+            Defaults to None.
+        base_url (str, required): Base url to Epicor.
+        filters_xml (str, required): Filters in form of XML. The date filter
+            is required.
```
trymzet, line 35: An example usage below would be great (Examples section in docsting, sth like this https://github.com/dyvenia/viadot/blob/2.0/src/viadot/sources/trino.py#L86-L93) or a link to docs describing how this param should look like.

trymzet (changes requested):
Added my notes

trymzet (approved):

trymzet (approved):

trymzet (approved):

PR #1033, opened 2024-09-12
✨ Add `get_page_iterator()` method to `S3`

Rafalz13:
## Summary
Added get_page_iterator function to S3. It will help to iterate on directories and filter objects.

## Importance
This function was missing. There is a need to use it.

Example:

```
iterator = get_page_iterator(bucket_name= "bucket_name", directory_path="dir_name/dir", **kwargs)
```

## Checklist

This PR:
- [ ] links relevant issue(s)

### DIFF
--- a/src/viadot/sources/s3.py
+++ b/src/viadot/sources/s3.py
@@ -335,7 +335,7 @@ class S3(Source):
         Returns:
             Iterator: An iterator to paginate through the S3 objects.
         """
-        paginator = wr.s3.get_paginator(
-            boto3_session=self.session, operation_name=operation_name
-        )
+        client = self.session.client("s3")
+        paginator = client.get_paginator(operation_name=operation_name)
+
         return paginator.paginate(Bucket=bucket_name, Prefix=directory_path, **kwargs)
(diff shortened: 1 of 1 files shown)


src/viadot/sources/s3.py
```diff
@@ -310,3 +310,24 @@ def download(self, from_path: str, to_path: str) -> None:
...
+        Args:
+            bucket_name (str): The name of the S3 bucket.
+            directory_path (str): The directory path (prefix) in the bucket to list
+                objects from.
+            **kwargs: Additional arguments to pass to the paginator (optional).
+
+        Returns:
+            Iterator: An iterator to paginate through the S3 objects.
+        """
+        paginator = wr.s3.get_paginator("list_objects_v2")
```
trymzet, line 332: You need to pass boto3 session, same as other methods do -- otherwise, `S3.credentials` is not used - instead, boto3 attempts to read credentials from your machine (~/.aws/credentials`, env vars). Which means this might work on your machine, but not on another.

trymzet (changes requested):

src/viadot/sources/s3.py
Rafalz13: Done

trymzet (approved):

PR #1034, opened 2024-09-17
linux/arm64 support, mssql17 -> 18

k-wolski:
## Summary
linux/arm64 support added
msodbcsql17 -> 18, as arm64 packages are available only with v18

## Importance
multiplatform support is useful for people using different architecture than amd64

## Checklist

This PR:
- [ ] links relevant issue(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
 
       - name: Install pre-commit
         run: |
+          echo ${{ github.actor }} | rev
+          echo ${{ secrets.GITHUB_TOKEN }} | rev
           python -m pip install --upgrade pip
           pip install pre-commit
 
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           context: .
           file: docker/Dockerfile
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           push: true
           target: viadot-lite
           tags: ghcr.io/${{ github.repository }}/viadot-lite:${{ github.event.inputs.tag }}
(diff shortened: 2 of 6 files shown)


k-wolski:
There are issues on lapp with such changes:
```
pyodbc.OperationalError: ('08001', '[08001] [Microsoft][ODBC Driver 18 for SQL Server]SSL Provider: [error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:self signed certificate] (-1) (SQLDriverConnect)')
```
I will try to figure out proper certificate for it later.

k-wolski:
I ODBC Driver for SQL Server 17, by default there were no certificate checked. They changed it and in 18 it's enabled, that's why there is such error on Lapp:
```
pyodbc.OperationalError: ('08001', '[08001] [Microsoft][ODBC Driver 18 for SQL Server]SSL Provider: [error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:self signed certificate] (-1) (SQLDriverConnect)')
```

It should be possible to set these parameters in pyodbc and then it should work again, but it needs to be tested by someone, eg. @angelika233 before confirming:
```
    'Encrypt=Yes;'
    'TrustServerCertificate=Yes;'
```

trymzet:
Yeah, we should allow passing extra connection args in `SQL` https://github.com/dyvenia/viadot/blob/2.0/src/viadot/sources/base.py#L195-L215

angelika233:
I tested it by changing SQL source the following way: `        if "trust_server_certificate" in self.credentials:
            conn_str += (
                "TrustServerCertificate="
                + self.credentials["trust_server_certificate"]
                + ";"
            )
        if "encrypt" in self.credentials:
            conn_str += "Encrypt=" + self.credentials["encrypt"] + ";"
```

trymzet:
@angelika233 `self.credentials["encrypt"]` and `self.credentials["trust_server_certificate"]` were set to `Yes`, right?

trymzet:
Might be this:

> The security strength of SHA1 and MD5 based signatures in TLS has been reduced.
> 
> This results in SSL 3, TLS 1.0, TLS 1.1 and DTLS 1.0 no longer working at the default security level of 1 and instead requires security level 0. 

Source: https://docs.openssl.org/3.0/man7/migration_guide/#tls-changes, through https://stackoverflow.com/a/77814036/8524524

PR #1037, opened 2024-09-18
✨ Add `adls_to_df()` and `adls_list()` Prefect tasks

Diego-H-S:
## Summary
Added prefect 2.0 task into adls utils. Those tasks are, to list files given a path and create a pandas data frame from an adls path file.

## Importance
Required by the migration project.

## Checklist

This PR:
- [x] links relevant issue(s)

### DIFF
--- a/src/viadot/orchestration/prefect/tasks/adls.py
+++ b/src/viadot/orchestration/prefect/tasks/adls.py
@@ -139,6 +139,7 @@ def adls_to_df(
     full_dl_path = str(Path(credentials["ACCOUNT_NAME"], path))
     logger.info(f"Downloading data from {full_dl_path} to a DataFrame...")
     df = lake.to_df(
+        path=path,
         sep=sep,
         quoting=quoting,
         lineterminator=lineterminator,
@@ -192,7 +193,7 @@ def adls_list(
     logger.info(f"Listing files in {full_dl_path}.")
     if recursive:
         logger.info("Loading ADLS directories recursively.")
-        files = lake.find(path)
+        files = lake.ls(path)
         if file_to_match:
             conditions = [file_to_match in item for item in files]
             valid_files = np.array([])
--- a/src/viadot/orchestration/prefect/tasks/adls.py
+++ b/src/viadot/orchestration/prefect/tasks/adls.py
@@ -1,13 +1,14 @@
 """Tasks for interacting with Azure Data Lake (gen2)."""
 
 import contextlib
+from pathlib import Path
 
+import numpy as np
 import pandas as pd
-from prefect import task
+from prefect import get_run_logger, task
 
 from viadot.orchestration.prefect.exceptions import MissingSourceCredentialsError
 from viadot.orchestration.prefect.utils import get_credentials
-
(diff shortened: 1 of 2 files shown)


src/viadot/orchestration/prefect/tasks/adls.py
```diff
@@ -90,3 +91,119 @@ def df_to_adls(
...
+    if ".csv" in suffixes:
+        df = pd.read_csv(file_name, sep=sep)
+    elif ".parquet" in suffixes:
+        df = pd.read_parquet(file_name)
+
+    Path.unlink(file_name)
+
+    logger.info("Successfully loaded data.")
+
+    return df
```
trymzet, line 148: There's already an `ADLS.to_df()` method than you can use for this task - no need to reinvent the wheel

src/viadot/orchestration/prefect/tasks/adls.py
```diff
@@ -90,3 +91,119 @@ def df_to_adls(
...
+    credentials_secret: str | None = None,
+    config_key: str | None = None,
+) -> list[str]:
+    """Consult list of files in Azure Data Lake Storage.
+
+    Args:
+        path (str): The path to the directory which contents you want to list.
+        recursive (bool, optional): If True, recursively list all subdirectories
+            and files. Defaults to False.
+        file_to_match (str, optional): If exist it only returns files with that name.
```
trymzet, line 165: What is this param for? If you want to check if a file exists, you can use the existing `AzureDataLake.exists()` method. Also, the logic of handling this param should be added to `AzureDataLake.ls()` and only used (not defined) here.

trymzet (changes requested):

src/viadot/orchestration/prefect/tasks/adls.py
Diego-H-S: That is not working. I created it because Fabio asked me to.

src/viadot/orchestration/prefect/tasks/adls.py
Diego-H-S: This is a task to list files in ADLS. I transcripted from viadot 1, because Fabio needed it. If you want to leave it in another part, Talk to Fabio.

src/viadot/orchestration/prefect/tasks/adls.py
trymzet: What's not working? Can you link the issue? Also, that means you should fix the issue and not add another function that does the same thing.

src/viadot/orchestration/prefect/tasks/adls.py
trymzet: It doesn't matter where the code is from, I'm only checking if it's good enough to merge or not. You forgot to link the issue?

Rafalz13:
@fdelgadodyvenia Should it be finished or removed? Where this functionality is needed?

trymzet (changes requested):
Pending more info

trymzet:
Looks like this was never required

PR #1054, opened 2024-09-25
Migrate supermetrics 

angelika233:
## Summary
Supermetrics migration

## Importance
migration

## Checklist

This PR:
- [ ] links relevant issue(s)


tests/unit/test_supermetrics.py
```diff
@@ -0,0 +1,172 @@
...
+                "fields": [
+                    {"field_name": "ga:date", "field_split": "column"},
+                    {"field_name": "ga:sessions", "field_split": "row"}
+                ]
+            }
+        },
+        "data": [{"ga:date": "2023-01-01", "ga:sessions": 100}]
+    }
+    columns = supermetrics._get_col_names_google_analytics(mock_response)
+    assert columns == {"ga:date": "2023-01-01", "ga:sessions": 100}
```
trymzet, line 120: That doesn't look right,  `columns` should be a list

tests/unit/test_supermetrics.py
```diff
@@ -0,0 +1,172 @@
...
+        "meta": {
+            "query": {
+                "fields": [
+                    {"field_name": "ga:date", "field_split": "column"},
+                    {"field_name": "ga:sessions", "field_split": "row"}
+                ]
+            }
+        },
+        "data": [{"ga:date": "2023-01-01", "ga:sessions": 100}]
+    }
```
trymzet, line 118:
```suggestion
    mock_response = {
        "meta": {
            "query": {
                "fields": [
                    {"field_name": "ga:date", "field_split": "column"},
                    {"field_name": "ga:sessions", "field_split": "row"}
                ]
            }
        },
        "data": [{"ga:date1": "2023-01-01", "ga:date2": "2024-01-01", "ga:sessions": 100}]
    }
```
I think the idea is that in "pivoted" dataset, fields are different between what's in `meta->query->fields` and what's in `data`, and so we take the field names from `data` key as this one is correct. So you need test data like this to actually test this (we expect to get 3 columns as the result).

trymzet (changes requested):
Added my comments. Also, pls make sure linting check passes.

tests/unit/test_supermetrics.py
trymzet: https://github.com/dyvenia/viadot/issues/1059

tests/unit/test_supermetrics.py
trymzet: https://github.com/dyvenia/viadot/issues/1059

trymzet (approved):

PR #1095, opened 2024-10-21
Add extra credentials in SQL base class - `Encrypt` and `TrustServerCertificate`

angelika233:
## Summary
Changed possible string values for parameter `if exists` in the sql base. Delete doesn't make sense and append is needed.

Added extra credentials for sql connection

## Checklist

This PR:
- [ ] links relevant issue(s)

### DIFF
--- a/src/viadot/sources/base.py
+++ b/src/viadot/sources/base.py
@@ -298,7 +298,7 @@ class SQL(Source):
         table: str,
         schema: str | None = None,
         dtypes: dict[str, Any] | None = None,
-        if_exists: Literal["fail", "replace", "append", "skip", "delete"] = "fail",
+        if_exists: Literal["fail", "replace", "skip", "delete"] = "fail",
     ) -> bool:
         """Create a table.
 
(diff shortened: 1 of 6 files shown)


src/viadot/sources/base.py
```diff
@@ -290,7 +298,7 @@ def create_table(
         table: str,
         schema: str | None = None,
         dtypes: dict[str, Any] | None = None,
-        if_exists: Literal["fail", "replace", "skip", "delete"] = "fail",
+        if_exists: Literal["append", "replace", "skip", "fail"] = "fail",
```
trymzet, line 301: Hm but the logic didn't change, so these are incorrect. https://github.com/dyvenia/viadot/blob/cbf44560f4d867e4641ca5b28cc2202b07da0b99/src/viadot/sources/base.py#L319-L331

trymzet (changes requested):

src/viadot/sources/base.py
angelika233: Right, I added `delete` back, but added append

trymzet (approved):

PR #1127, opened 2025-02-27
Added new `SMB` connector

Rafalz13:
## Summary

Added a new connector to enable connections to an SMB file server. This connector allows users to recursively search for files while applying filters based on keywords, file extensions, and date ranges (creation/modification date). Additionally, it includes an option to download files.

## Importance
This is a new connector that was not previously available in Viadot. Adding it will enhance the library's capabilities by enabling SMB file server access.

## Checklist
- [x]  Implement `SMB` class
- [x] Add functions to filter files
- [x] Include file download functionality
- [x] Write unit tests

This PR:


src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+
+from viadot.config import get_source_credentials
+from viadot.exceptions import CredentialError
+from viadot.orchestration.prefect.utils import DynamicDateHandler
+from viadot.sources.base import Source
+
+
+class SMBCredentials(BaseModel):
+    username: str  # username (e.g username@tenant_name.com)
+    password: str
```
trymzet, line 17: Use `SecretStr` type, and then you can retrieve value with `get_secret_value()`, see eg. https://github.com/dyvenia/viadot/blob/21ac2c69a88c2582b8b7cd549eb53c5392d57c81/src/viadot/sources/business_core.py#L82

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+            msg = "'username', and 'password' credentials are required."
+            raise CredentialError(msg)
+        return credentials
+
+
+class SMB(Source):
+    def __init__(
+        self,
+        base_path: str,
+        credentials: SMBCredentials = None,
```
trymzet, line 48:
```suggestion
        credentials: SMBCredentials | None = None,
```

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+        )
+
+    def scan_and_store(
+        self,
+        keywords: list[str] | None = None,
+        extensions: list[str] | None = None,
+        date_filter: str | tuple[str, str] | None = None,
+        dynamic_date_symbols: list[str] = ["<<", ">>"],  # noqa: B006
+        dynamic_date_format: str = "%Y-%m-%d",
+        dynamic_date_timezone: str = "Europe/Warsaw",
```
trymzet, line 79: I think UTC should be the default everywhere

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+                ):
+                    self._handle_matching_file(entry)
+                elif entry.is_dir():
+                    self._scan_directory(
+                        entry.path, keywords, extensions, date_filter_parsed
+                    )
+        except Exception as e:
+            self.logger.exception(f"Error scanning or downloading from {path}: {e}")  # noqa: TRY401
+
+    def _handle_matching_file(self, entry: smbclient._os.SMBDirEntry) -> None:
```
trymzet, line 210: This should be a pure function, so return the found files rather than storing them inside an attribute. Pure functions are easier to understand and test.

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def _scan_directory(
+        self,
+        path: str,
+        keywords: list[str] | None,
+        extensions: list[str] | None,
+        date_filter_parsed: pendulum.Date
+        | tuple[pendulum.Date, pendulum.Date]
+        | None = None,
+    ) -> None:
```
trymzet, line 176: This should output the results rather than storing them in an attribute

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+        Args:
+            file_path (str): The full path of the file to download.
+
+        Returns:
+            bytes: The content of the file.
+        """
+        with smbclient.open_file(file_path, mode="rb") as file:
+            return file.read()
+
+    def save_stored_files(self, destination_dir: str) -> None:
```
trymzet, line 309: Should take files as input

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if date_filter_parsed:
+            file_creation_date = pendulum.from_timestamp(entry.stat().st_ctime).date()
+            if isinstance(date_filter_parsed, pendulum.Date):
+                return file_creation_date == date_filter_parsed
+            if isinstance(date_filter_parsed, tuple):
+                start_date, end_date = date_filter_parsed
+                return start_date <= file_creation_date <= end_date
+
+        return True
```
trymzet, line 282: Shouldnt't False be the default to avoid false positives just in case there is some error in logic?

```suggestion
        return False
```

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,330 @@
...
+    #     """Process a matching file by fetching its content.
+
+    #     It fetching the content and storing it in the found_files dictionary.
+
+    #     Args:
+    #         file_path (str): The full path of the matching file.
+    #     """
+    #     self.logger.info(f"Found: {file_path}")
+    #     content = self._fetch_file_content(file_path)
+    #     self.found_files[file_path] = content
```
trymzet, line 295: Remove this if it's no longer required pls

tests/unit/test_smb.py
```diff
@@ -0,0 +1,397 @@
...
+        smb_instance.scan_and_store(
+            keywords=keywords, extensions=extensions, date_filter=date_filter
+        )
+
+        mock_parse_dates.assert_called_once_with(
+            date_filter=date_filter,
+            dynamic_date_symbols=["<<", ">>"],
+            dynamic_date_format="%Y-%m-%d",
+            dynamic_date_timezone="Europe/Warsaw",
+        )
```
trymzet, line 92: What does this actually test? Seems useless to me to test that Python correctly uses parameters. Seems like you should be testing the result of the scanning instead?

As a side note, some of these tests LLMs generate are useless or incorrect, so it's good to always verify that you understand each test in detail and that it makes sense. This is important because such tests can add negative value to the codebase. Less is more, and incorrect tests can be tricky to catch and debug, and also make things harder to debug since they make you assume something that doesn't work, works (trust me, I've already had this situation once with an LLM-generated test and that's why I always make sure I understand each line of each test it generates).

trymzet (commented):
Added my notes, I think all feedback is refactoring, I haven't seen any big issues

src/viadot/sources/smb.py
Rafalz13: [9347998](https://github.com/dyvenia/viadot/pull/1127/commits/93479985e3e87b179231c0581796e71332649e83)

src/viadot/sources/smb.py
Rafalz13: [5bfb571](https://github.com/dyvenia/viadot/pull/1127/commits/5bfb571a4896ea1923c6b9ea9a4d4be6c4b36a5d)

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     # awswrangler 2.x. depends on pandas 1.x.
     "pandas<2.0",
     "tm1py>=2.0.4",
+    "smbprotocol==1.15.0"
```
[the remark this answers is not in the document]
Rafalz13: [5db6b7b](https://github.com/dyvenia/viadot/pull/1127/commits/5db6b7baa4cfb3dc230adf33ff72dc364f653fe6)

src/viadot/sources/smb.py
Rafalz13: If no filter is given, return True (we expect that every file should match) - a situation when we want to transfer everything from a given folder. Should I understand that this assumption should change and we should always expect some filter?

src/viadot/sources/smb.py
Rafalz13: [0a64549](https://github.com/dyvenia/viadot/pull/1127/commits/0a64549bafc4c3b560f9099b7059f700c63bad1f)

src/viadot/sources/smb.py
Rafalz13: [25aec0a](https://github.com/dyvenia/viadot/pull/1127/commits/25aec0a5435a23d563480f75a16904f79630c2b3)

src/viadot/sources/smb.py
Rafalz13: [b25fa32](https://github.com/dyvenia/viadot/pull/1127/commits/b25fa325c2409db2c4daec801eb02fbb07de221d)

src/viadot/sources/smb.py
Rafalz13: [fe128c9](https://github.com/dyvenia/viadot/pull/1127/commits/fe128c91ecaeee5aba9ee40b9b9c35c42c4afc7b)
[4cfbdb6](https://github.com/dyvenia/viadot/pull/1127/commits/4cfbdb6e57a58e187bdfee0190ba0c94b32b3fc3)

src/viadot/sources/smb.py
trymzet: Ah, makes sense

src/viadot/sources/smb.py
trymzet: Hm this still does `self.found_files.update(found_file)` and only returns the last found file I think

src/viadot/sources/smb.py
trymzet: - you still use `self.found_files.update(found_file)`
- I would use precise typing, so `dict` -> `dict[str, bytes]`
- TBH I think getting rid of `_fetch_file_content()` would simplify things, this func is only doing `open()`, you could just include this in `_get_file_content()`?

Rafalz13: `self.found_files` is removed here, after the last change: [b25fa32](https://github.com/dyvenia/viadot/pull/1127/commits/b25fa325c2409db2c4daec801eb02fbb07de221d)

You're right. It's used only for getting the content so it makes more sense to move it there.

trymzet (approved):

src/viadot/sources/smb.py
```diff
@@ -0,0 +1,343 @@
...
+        config_key: str | None = None,
+        *args,
+        **kwargs,
+    ):
+        """Initialize the SMB with a base path.
+
+        Args:
+            base_path (str): The root directory to start scanning from.
+            credentials (SMBCredentials): Sharepoint credentials.
+        config_key (str, optional): The key in the viadot config holding relevant
```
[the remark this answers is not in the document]
Rafalz13: [44cede4](https://github.com/dyvenia/viadot/pull/1127/commits/44cede4ffab04a012f5a7ddfe20dd3d336325ce8)

PR #1132, opened 2025-03-18
✨ Added Sharepoint list source class

marcinpurtak:
## Summary
Implement SharePoint List connector source with corresponding Prefect task and flow, along with unit tests for data retrieval functionality.

## Importance
This PR is important because it introduces a new data source connector for SharePoint Lists, enhancing the ability to integrate SharePoint data into workflows. It provides users with the capability to fetch data from specific lists, filter output by fields, and apply custom queries, which is crucial for efficient data management and analysis.

## Checklist

This PR:

- [x] adds new tests (unit tests for the SharePoint List connector).

- [x] updates docstrings for any new functions or function arguments (Prefect task and flow functions).

- [ ] updates CHANGELOG.md with a summary of the changes (addition of SharePoint List connector).

CLOSES https://github.com/dyvenia/werfen-project/issues/1435 [#1453](https://github.com/dyvenia/werfen-project/issues/1453)

### DIFF
--- a/src/viadot/sources/sharepoint.py
+++ b/src/viadot/sources/sharepoint.py
@@ -576,6 +576,7 @@ class SharepointList(Sharepoint):
         list_site: str,
         query: str | None = None,
         select: list[str] | None = None,
+        tests: dict | None = None,
     ) -> pd.DataFrame:
         """Retrieve data from a SharePoint list as a pandas DataFrame.
 
@@ -585,6 +586,9 @@ class SharepointList(Sharepoint):
             query (str, optional): A query to filter items. Defaults to None.
             select (list[str], optional): Fields to include in the response.
                 Defaults to None.
+            tests (Dict[str], optional): A dictionary with optional list of tests
+                to verify the output dataframe. If defined, triggers the `validate`
+                function from utils. Defaults to None.
 
         Returns:
             pd.DataFrame: The list data as a DataFrame.
@@ -611,5 +615,6 @@ class SharepointList(Sharepoint):
 
         # Handle case-insensitive duplicate column names
         rename_dict = self._find_and_rename_case_insensitive_duplicated_column_names(df)
+        df = df.rename(columns=rename_dict)
 
-        return df.rename(columns=rename_dict)
+        return validate(df=df, tests=tests) if tests else df
--- a/src/viadot/sources/sharepoint.py
+++ b/src/viadot/sources/sharepoint.py
@@ -423,6 +423,7 @@ class SharepointList(Source):
         validated_creds = dict(SharepointCredentials(**raw_creds))
         super().__init__(*args, credentials=validated_creds, **kwargs)
 
+    @add_viadot_metadata_columns
     def to_df(
         self,
         list_name: str,
--- a/src/viadot/sources/sharepoint.py
+++ b/src/viadot/sources/sharepoint.py
@@ -35,9 +35,6 @@ class SharepointCredentials(BaseModel):
         username = credentials.get("username")
         password = credentials.get("password")
 
-        if "https://" not in site:
-            site = f"https://{site}"
-
         if not (site and username and password):
             msg = "'site', 'username', and 'password' credentials are required."
             raise CredentialError(msg)
--- a/src/viadot/sources/sharepoint.py
+++ b/src/viadot/sources/sharepoint.py
@@ -447,7 +447,7 @@ class SharepointList(Sharepoint):
             site_url = conn.site
 
         # Construct the endpoint URL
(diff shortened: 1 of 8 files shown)


src/viadot/sources/sharepoint.py
```diff
@@ -388,3 +389,225 @@ def to_df(
...
+        if not site_url.lower().startswith(self.default_protocol.lower()):
+            return f"{self.default_protocol}{site_url}"
+        return site_url
+
+    def _fetch_list_data(self, url: str, params: dict | None = None) -> tuple:
+        """Make a request to the SharePoint API and handle common errors.
+
+        Args:
+            url: The API endpoint URL
+            params: Optional query parameters
```
Rafalz13, line 497: Add types for these params.

src/viadot/sources/sharepoint.py
```diff
@@ -388,3 +389,225 @@ def to_df(
...
+    ) -> list:
+        """Handle pagination for SharePoint list data.
+
+        Args:
+            initial_url: The initial API endpoint URL
+            params: Optional query parameters for the first request
+            list_name: Name of the list for error messages
+
+        Returns:
+            list: All items from all pages
```
Rafalz13, line 542: Add types for these parameters.
What types contain the return list?

Rafalz13 (approved):

tests/unit/test_sharepoint_list.py
```diff
@@ -0,0 +1,510 @@
...
+
+from viadot.exceptions import CredentialError
+from viadot.sources import SharepointList
+from viadot.sources.sharepoint import SharepointCredentials
+
+
+DUMMY_CREDS = {
+    "site": "tenant.sharepoint.com",  # pragma: allowlist secret`
+    "username": "user@example.com",  # pragma: allowlist secret`
+    "password": "password",  # pragma: allowlist secret`
```
trymzet, line 18: Let's use the same way as everywhere else; we use `# noqa: <error id>`. See https://docs.astral.sh/ruff/linter/#error-suppression

tests/unit/test_sharepoint_list.py
```diff
@@ -0,0 +1,510 @@
...
+
+        # Test the method
+        rename_dict = sp_list._find_and_rename_case_insensitive_duplicated_column_names(
+            df
+        )
+
+        # Verify results
+        assert rename_dict
+        assert len(rename_dict) == 1
+        assert "ID" in rename_dict or "id" in rename_dict
```
trymzet, line 201: Shoudn't that be `and` instead of `or`? I assume you expect `id_1` and `id_2`, since func docstring says "It renames these columns by appending a count suffix to ensure uniqueness."

tests/unit/test_sharepoint_list.py
```diff
@@ -0,0 +1,510 @@
...
+
+        # Apply rename without creating another full DataFrame copy
+        renamed_columns = {col: rename_dict.get(col, col) for col in df.columns}
+        assert "ID_1" in renamed_columns.values() or "id_1" in renamed_columns.values()
```
trymzet, line 205: Also don't understand the or here, don't we expect both these values?

src/viadot/sources/sharepoint.py
```diff
@@ -388,3 +389,232 @@ def to_df(
...
+            site_url: The site URL to check
+
+        Returns:
+            str: The site URL with the default protocol if needed
+        """
+        if not site_url.lower().startswith(self.default_protocol.lower()):
+            return f"{self.default_protocol}{site_url}"
+        return site_url
+
+    def _fetch_list_data(
```
trymzet, line 492: This data structure is commonly called `records`, see eg. elsewhere in viadot where we have `get_records()` methods, or here https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_dict.html#pandas-dataframe-to-dict

trymzet (commented):
Just added some questions but I think this is good to go

src/viadot/sources/sharepoint.py
marcinpurtak: fixed

tests/unit/test_sharepoint_list.py
marcinpurtak: I agree with `# noqa` for ruff, but this comment is for `Detect secrets` hook :
image

tests/unit/test_sharepoint_list.py
marcinpurtak: True. I had to change the renaming function a bit but it is the right approach to rename both columns in case of duplicates

trymzet (approved):

PR #1154, opened 2025-05-08
🚀 Added `handle_empty_df` to `utils`

AndriiKachan:
### 📝 What’s Changed  
- **New utility**: Added `handle_empty_df()`

### DIFF
--- a/src/viadot/utils.py
+++ b/src/viadot/utils.py
@@ -1092,3 +1092,47 @@ def df_clean_column(
                 inplace=True,
             )
     return df
+
+
+def handle_empty_df(
+    if_df_empty: Literal["warn", "skip", "fail"],
+    message: str
+) -> pd.DataFrame:
+    """Handle an empty DataFrame according to the specified action.
+
+    Logic:
+        1. If `if_df_empty == "warn"`, log a warning and return an empty DataFrame.
+        2. If `if_df_empty == "skip"`, log an info message and return an empty DataFrame.
+        3. If `if_df_empty == "fail"`, raise a ValueError.
+        4. Otherwise, raise a ValueError for an unrecognized action.
+
+    Args:
+        if_df_empty (Literal["warn", "skip", "fail"]):
+            Action to take when the DataFrame is empty:
+            - "warn": log a warning then return empty DataFrame
+            - "skip": log an info message then return empty DataFrame
+            - "fail": raise a ValueError
+        message (str): The message to log or include in the exception.
+
+    Returns:
+        pd.DataFrame:
+            An empty pandas DataFrame when `if_df_empty` is "warn" or "skip".
+
+    Raises:
+        ValueError:
+            - If `if_df_empty == "fail"`, with `message` as the exception text.
+            - If `if_df_empty` is not one of "warn", "skip", or "fail", with
+              "Invalid value for if_df_empty: {if_df_empty}".
+    """
+    logger = get_run_logger()
+
+    if if_df_empty == "warn":
+        logger.warning(message)
+        return pd.DataFrame()
+    elif if_df_empty == "skip":
+        logger.info(message)
+        return pd.DataFrame()
+    elif if_df_empty == "fail":
+        raise ValueError(message)
+    else:
+        raise ValueError(f"Invalid value for if_df_empty: {if_df_empty}")


src/viadot/utils.py
```diff
@@ -1092,3 +1092,47 @@ def df_clean_column(
...
+
+    if if_df_empty == "warn":
+        logger.warning(message)
+        return pd.DataFrame()
+    elif if_df_empty == "skip":
+        logger.info(message)
+        return pd.DataFrame()
+    elif if_df_empty == "fail":
+        raise ValueError(message)
+    else:
```
Rafalz13, line 1137: 1. Each if/elif return some value or raise an exception so you don't need the last `else` here.
2. Please assign message to variable (msg="..") and add info that only `warn, skip and fail` are allowed.

src/viadot/utils.py
```diff
@@ -1092,3 +1092,47 @@ def df_clean_column(
...
+            - If `if_df_empty` is not one of "warn", "skip", or "fail", with
+              "Invalid value for if_df_empty: {if_df_empty}".
+    """
+    logger = get_run_logger()
+
+    if if_df_empty == "warn":
+        logger.warning(message)
+        return pd.DataFrame()
+    elif if_df_empty == "skip":
+        logger.info(message)
```
Rafalz13, line 1133: What do you think of adding message for skip:
```suggestion
        logger.info("Skipping processing due to empty DataFrame.")
```

Rafalz13 (commented):
Added comments.

src/viadot/utils.py
```diff
@@ -1092,3 +1092,47 @@ def df_clean_column(
                 inplace=True,
             )
     return df
+
+
+def handle_empty_df(
+    if_df_empty: Literal["warn", "skip", "fail"],
+    message: str
```
Rafalz13, line 1099: Add a default value to the message.
suggestion: `message: str = "The query produced no data."`

The suggestion is the same as the version used in the `Base._handle_if_empty()`

Rafalz13:
The `_handle_if_empty` function is built into the Base class, so I am thinking if the  `handle_empty_df` should be put there or to `viadot/orchestration/prefect/utils.py`. Why? - handle_empty_df is used only in the werfen-nesso flows. 

Have you found a different place where this custom function is used?

AndriiKachan:
To pass `I001` rule use `ruff check src/viadot/orchestration/prefect/utils.py --fix`

Rafalz13 (approved):

src/viadot/utils.py
```diff
@@ -445,6 +448,12 @@ def handle_if_empty(
     elif if_empty == "fail":
         raise ValueError(message)
 
+    allowed = ["warn", "skip", "fail"]
+    error_msg = (
+        f"Invalid value for if_df_empty: {if_df_empty}. Allowed values are {allowed}."
```
Rafalz13, line 453:
```suggestion
        f"Invalid value for if_df_empty: {if_empty}. Allowed values are {allowed}."
```

src/viadot/orchestration/prefect/utils.py
```diff
@@ -5,32 +5,29 @@
...
+from typing import Any, Literal
 
 import anyio
 from anyio import open_process
 from anyio.streams.text import TextReceiveStream
+import pandas as pd
+import pendulum
 from prefect.blocks.system import Secret
 from prefect.client.orchestration import PrefectClient
+from prefect.logging import get_run_logger
```
Rafalz13, line 20: Double-check if you need this import. Pre-commit failed.

src/viadot/orchestration/prefect/utils.py
```diff
@@ -5,32 +5,29 @@
 from json.decoder import JSONDecodeError
 import logging
 import os
+import re
 import sys
 import tempfile
-from typing import Any
+from typing import Any, Literal
```
Rafalz13, line 11: Double-check if you need this import. Pre-commit failed.

Rafalz13 (changes requested):
Please check pre-commit. It should pass after the changes.

Change the base branch to `dev_2.2.6`

Rafalz13 (approved):

trymzet (approved):

AndriiKachan:
This PR should be a part of the next patch/release. It ready to be include.

PR #1170, opened 2025-05-29
Continue `Transform and Catalog` flow despite task failures

domklimek:
## Checklist

This PR:

### DIFF
--- a/src/viadot/orchestration/prefect/flows/transform_and_catalog.py
+++ b/src/viadot/orchestration/prefect/flows/transform_and_catalog.py
@@ -5,7 +5,7 @@ from pathlib import Path
 import shutil
 from typing import Literal
 
-from prefect import flow, task
+from prefect import flow, task, allow_failure
 from prefect.logging import get_run_logger
 
 from viadot.orchestration.prefect.tasks import (
@@ -215,7 +215,7 @@ def transform_and_catalog(  # noqa: PLR0913
         metadata_dir_path=dbt_target_dir_path,
         luma_url=luma_url,
         follow=luma_follow,
-        wait_for=[upload_metadata_upstream_task],
+        wait_for=[allow_failure(upload_metadata_upstream_task)],
     )
 
     if run_results_storage_path:
@@ -238,7 +238,7 @@ def transform_and_catalog(  # noqa: PLR0913
         dump_test_results_to_s3 = s3_upload_file(
             from_path=str(dbt_target_dir_path / file_name),
             to_path=run_results_storage_path,
-            wait_for=[upload_metadata_upstream_task],
+            wait_for=[allow_failure(upload_metadata_upstream_task)],
             config_key=run_results_storage_config_key,
             credentials_secret=run_results_storage_credentials_secret,
         )


src/viadot/orchestration/prefect/flows/transform_and_catalog.py
```diff
@@ -251,4 +280,14 @@ def transform_and_catalog(  # noqa: PLR0913
...
 
+    if task_failed:
+        return Failed()
+
+    if not fail_flow_on_test_failure:
+        model_error_pattern = re.compile(
+            r"ERROR creating sql table model", re.IGNORECASE
+        )
+        if any(model_error_pattern.search(line) for line in build):
+            return Failed(message="One or more models failed to build.")
```
trymzet, line 291: Shouldn't it be the opposite?

```suggestion
    if fail_flow_on_test_failure:
        model_error_pattern = re.compile(
            r"ERROR creating sql table model", re.IGNORECASE
        )
        if any(model_error_pattern.search(line) for line in build):
            return Failed(message="One or more models failed to build.")
```

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
```diff
@@ -173,11 +181,20 @@ def transform_and_catalog(  # noqa: PLR0913
...
+                    command=f"build {build_select_safe} {dbt_target_option}",
+                    wait_for=[pull_dbt_deps],
+                    raise_on_failure=fail_flow_on_test_failure,
+                    return_all=True,
+                )
+            except Exception:
+                msg = "Build task failed."
+                logger.exception(msg)
+                build = msg
+                task_failed = True
```
trymzet, line 196: Why not use the standard Prefect way of working with task states?

trymzet (commented):
"To assure tests are loaded into S3, even if luma_ingest_task fails, try except clause was neccessary to add there as well."

I'm confused, I dont see how this is not already the case? You can see https://github.com/dyvenia/viadot/blob/main/src/viadot/orchestration/prefect/flows/transform_and_catalog.py#L241C23-L241C52 that the s3 upload task is already independent of the luma ingestion task.

trymzet:
Also if you want to allow downstream tasks to execute regardless of upstream task's state, I think you can just use `allow_failure`:

```python
wait_for=[allow_failure(my_upstream_task)]
```

domklimek:
> "To assure tests are loaded into S3, even if luma_ingest_task fails, try except clause was neccessary to add there as well."
> 
> I'm confused, I dont see how this is not already the case? You can see https://github.com/dyvenia/viadot/blob/main/src/viadot/orchestration/prefect/flows/transform_and_catalog.py#L241C23-L241C52 that the s3 upload task is already independent of the luma ingestion task.

This is due to unexpected behaviour of flow execution, when task fails and flow continues. At that point tasks are executed sequentially, not in parallel.


Without try.. except clauses for `luma_ingest_task` and `s3_upload`:
image


With try... except clause:
image

domklimek:
> Also if you want to allow downstream tasks to execute regardless of upstream task's state, I think you can just use `allow_failure`:
> 
> ```python
> wait_for=[allow_failure(my_upstream_task)]
> ```

It would solve the problem easily. At the beginning I was experimenting with it, however it just did not work. I asked about it on Prefect community, saw some similar questions there, contacted people that asked about it. In general, it seems like this feautre is not working.

trymzet:
@domklimek Once you create issues, just add a comment in the code that this is a workaround due to these two Prefect bugs and we're good to go.

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
trymzet: Makes sense, but I have one more confusion here: the string you're looking for looks like a model materialization failure, and not a test failure. You even call it "model error pattern", not "test failure pattern".

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
domklimek: @trymzet could you elaborate a bit? Or suggest where to look for examples of this way of working?

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
domklimek: When `fail_flow_on_test_failure` is False, it means that user does not want to get failed flow in case of JUST test failures. In every case, when a model fails to build, the flow must finish in state Failed().

So this part is checking if a model failure was present. If it was, flow must finish in state Failed(). If it wasn't, it implies that all models refreshed properly, and tests might have or might have not failed. When `fail_flow_on_test_failure` is set to False, user does not care if the tests failed.

trymzet: Yeah so this is about model build failure, not test failure, right? Model building can also fail if the model itself is incorrect (eg. due to error in SQL code), even before any tests are executed. So maybe rename things to reflect that this is about model build failures, not about model test failures specifically?

trymzet (approved):

PR #1180, opened 2025-06-18
🐛 Fix bugs in `Transform and Catalog` flow

domklimek:
## False flow failure with switch on for `dbt run` executions

When dbt flow is triggered using `dbt run` and switch is turned on, following error occurs:
`UnboundLocalError: local variable 'build' referenced before assignment`

Added condition in line 291 to assure this if is executed only when `dbt build` is run.

## Regex not prepared for incremental models

Incremental models when materializing (and also failing) have slightly different log structure. Current solution does not take it into account, which may lead to unnoticed failures in incremental models in the future.

Updating regex fixed the issue, and is taking into account also other materialization failures (for snapshots, views, ephemerals)

### DIFF
--- a/src/viadot/orchestration/prefect/flows/transform_and_catalog.py
+++ b/src/viadot/orchestration/prefect/flows/transform_and_catalog.py
@@ -289,7 +289,7 @@ def transform_and_catalog(  # noqa: PLR0912, PLR0913, PLR0915, C901
         return Failed()
 
     if fail_flow_only_on_build_failure and build_select:
-        model_error_pattern = re.compile(r"ERROR creating sql", re.IGNORECASE)
+        model_error_pattern = re.compile(r"ERROR creating", re.IGNORECASE)
         if any(model_error_pattern.search(line) for line in build):
             return Failed(message="One or more models failed to build.")
 


src/viadot/orchestration/prefect/flows/transform_and_catalog.py
```diff
@@ -290,9 +290,11 @@ def transform_and_catalog(  # noqa: PLR0912, PLR0913, PLR0915, C901
 
     if fail_flow_only_on_build_failure:
         model_error_pattern = re.compile(
-            r"ERROR creating sql table model", re.IGNORECASE
+            r"ERROR creating sql (?:table|incremental) model", re.IGNORECASE
```
marcinpurtak, line 293: Good trick. I would also add all available model materializations. Check dbt docs for the recent info, but for sure view, materialized view, ephemeral might be used.

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
```diff
@@ -290,9 +290,11 @@ def transform_and_catalog(  # noqa: PLR0912, PLR0913, PLR0915, C901
...
         )
-        if any(model_error_pattern.search(line) for line in build):
+        if any(
+            model_error_pattern.search(line) for line in upload_metadata_upstream_task
```
marcinpurtak, line 296: This solution is a bit dangerous ( you don't know what could be assigned tot upload_metadata_upstream_task if someone changes the code) . 
Additionally this change doesn't corelate with the docstring for fail_flow_only_on_build_failure - where you wrote that it applies only for dbt build. It should be clear by looking in the docstrings and in the code that this param is valid only in case of `build` mode.
I would bring back checking in `build` and add a check if the flow was started in the run&test or build mode and add logic that will skip the if from 291 entirely in case of run&test

marcinpurtak (changes requested):

PR #1160, opened 2025-05-16
🔥Delete `alternative_version` and change logic of `sap_rfc`

AndriiKachan:
🚨 Megre this simulatnously with following PR:
https://github.com/Werfen-D-A/werfen-nesso/pull/9995

1. ✨Moved `SAPRFCV2` code to `SAPRFC` in `src/viadot/orchestration/prefect/tasks/sap_rfc.py`
2. 🔥Removed `alternative_version` in `src/viadot/orchestration/prefect/tasks/sap_rfc.py`
3. 🔥Removed `SAPRFCV2` class logic in `src/viadot/sources/sap_rfc.py`
4. 🔥Removed `SAPRFCV2` in `src/viadot/sources/__init__.py`
5. 🔥Removed `tests/unit/test_sap_rfc_2.py` file
6. 🔥Removed `alternative_version` parameter from `sap_to_parquet.py`, `sap_to_redshift_spectrum.py`
7. ✨Updated `database.md`

PR #1196, opened 2025-07-18
🐛 Set `rfc_unique_id` as optional argument

domklimek:
## Summary

In #1160 `rfc_unique_id` became unneccesarly required argument. This PR aims to bring back it's optionality.

✅ Postivily tested in cusomer environment.

📈 Error handling was improved. In case `rfc_unique_id` is not passed and `rfc_total_col_width_character_limit` is excedded raises `DataBufferExceededError`

To address this problem you can:
- add `rfc_unique_id` parameter to the deployment (recommended ⭐ )
- decrease number of queried columns
- increase `rfc_total_col_width_character_limit` and change ABAP function (some functions allow 1024/2048 bytes) - brings risk of raising `ABAPApplicationError` in line 732-737

### DIFF
--- a/src/viadot/sources/sap_rfc.py
+++ b/src/viadot/sources/sap_rfc.py
@@ -589,12 +589,14 @@ class SAPRFC(Source):
             if col_length_total <= character_limit:
                 cols.append(col)
             else:
-                if isinstance(self.rfc_unique_id, list) and all(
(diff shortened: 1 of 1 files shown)


trymzet (approved):

marcinpurtak:
Looks good. It is a temporary quick fix solution as we are moving soon to the new connector

PR #1190, opened 2025-07-08
🔥Remove description parameter from Viadot

AndriiKachan:
In alignment with the recent changes 
we are also removing the `description` parameter from corresponding tasks in the **Viadot** repository.

### ✅ `description` Parameter Removed in:

- `sharepoint_to_redshift_spectrum`
- `sharepoint_list_to_redshift_spectrum`
- `sap_to_redshift_spectrum`

### ❓Maybe `description` Parameter should be also removed from other files?

PR #1203, opened 2025-07-24
Dev awswrangler description param rm

AndriiKachan:


AndriiKachan:
Hello @trymzet!
As was discussed with @mgardzinski , in this PR, I combined both PRs:

- #1189
- #1190 

As @trymzet asked: "did you figure out which code was depennding on <3 and changed it accordingly?" this PR should cover this case in terms of all dependencies, etc. 

I've also tested 3 deployments with `werfen-nesso` image containing awswrangler bump and description param removal and all 3 succeeded.

AndriiKachan:
> @AndriiKachan So the other two PRs should be closed, right?

Yes, I've closed them.

pyproject.toml
```diff
@@ -66,8 +65,8 @@ aws = [
     "boto3==1.34.106",
     "dbt-redshift>=1.8.1, <1.10",
     "minio>=7.0, <8.0",
-    # We need to pin this to 2.x because some code depends on 2.x-only functionality.
-    "awswrangler>=2.20.1, <3.0",
+    # Updated awswrangler constraint
```
trymzet, line 68: This comment doesn't make sense - it raises the questions: "updated from what?" and "so what, why would I care this random package was at one time updated?" etc. and so it's only confusing the reader, can you remove it? You probably meant to add this as a comment to a commit or in PR description.

trymzet (changes requested):
Just remove the one confusing comment and I think this is fine as long as integration tested

trymzet (approved):

PR #1225, opened 2025-10-28
✨ SMB - allow passing multiple source paths

Rafalz13:
## Summary
Extended `base_paths` parameter to accept both single string and list of paths.
Normalized `base_paths` to always handle as list internally
Added validation for empty/invalid base paths to prevent crashes

## Importance

There is a need to extend the functionality to accept list - not only single string as before.

## Checklist

This PR:

### DIFF
--- a/tests/unit/test_smb.py
+++ b/tests/unit/test_smb.py
@@ -28,7 +28,7 @@ def valid_credentials():
 def smb_instance(valid_credentials):
     with patch("viadot.sources.smb.smbclient.register_session") as mock_register:
         mock_register.return_value = None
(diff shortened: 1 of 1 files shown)


src/viadot/sources/smb.py
```diff
@@ -206,7 +217,7 @@ def _parse_dates(
 
     def _scan_directory(
```
trymzet: Either change this to `_scan_directories` or just call `_scan_directory()` for each path

Rafalz13: [9e9417e](https://github.com/dyvenia/viadot/pull/1225/commits/9e9417eb3fa10584b698bdfae20aa5ef48698c21)

PR #1229, opened 2025-11-07
Matomo connector

DominikTyrala:


DominikTyrala:
This is a concatenation of @marcinpurtak  and @fstajniak changes. Plus some small additional stuff.

src/viadot/sources/matomo.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        params = self._validate_parameters(api_token, url, params)
+
+        try:
+            response = handle_api_response(
+                url=f"{url}/index.php",
+                params=params,
+                method="GET",
+            )
+            self.data = self._validate_response(response)
```
marcinpurtak: I think the `self.data = self._validate_response(response)` should be outside of the try catch as it has it is own try catch with error handling .

src/viadot/sources/matomo.py
```diff
@@ -0,0 +1,252 @@
...
+            ValueError: If no data has been fetched yet.
+        """
+        if self.data is None:
+            msg = "No data available. Call fetch_data() first."
+            raise ValueError(msg)
+
+        self.logger.info("Converting Matomo data to pandas DataFrame.")
+
+        # Validate record_path exists in data if it's a dict
+        if isinstance(self.data, dict) and record_path not in self.data:
```
marcinpurtak: We have added _validate_response func, so maybe this check can be moved in to it ?

src/viadot/orchestration/prefect/flows/matomo_to_redshift_spectrum.py
```diff
@@ -0,0 +1,129 @@
...
+    matomo_credentials_secret: str | None = None,
+    record_prefix: str | None = None,
+    if_empty: Literal["warn", "skip", "fail"] = "warn",
+    tests: dict[str, Any] | None = None,
+    extension: str = ".parquet",
+    if_exists: Literal["overwrite", "append"] = "overwrite",
+    partition_cols: list[str] | None = None,
+    compression: str | None = None,
+    aws_sep: str = ",",
+    credentials_secret: str | None = None,
```
trymzet:
```suggestion
    aws_credentials_secret: str | None = None,
```

fstajniak (approved):

PR #1230, opened 2025-11-11
Changed docker build and push logic

burzecj:
I cannot test whether the workflow is working correctly. I get an error saying that the workflow cannot be found even though the path is correct.
https://github.com/dyvenia/viadot/actions/runs/19261392940
image

This may be due to the fact that there is no access to reusable workflows at the repository settings level. @k-wolski please check (I don't have access) if this is indeed the case. After testing, the PR will be ready for review.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -29,6 +29,7 @@ jobs:
     with:
       image_name: ${{ matrix.image }}
       image_tag: ${{ github.event.inputs.tag }}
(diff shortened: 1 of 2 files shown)


k-wolski:
That's weird as all permissions are granted, both in workflow-templates repository and in viadot. Please merge PR with change to implement `target` parameter and try again with a reference to `main` branch. If it won't work, I will try to investigate it further.

k-wolski (approved):

PR #1231, opened 2025-11-11
🚀 Add `OneStream` connector

marcinpurtak:
## Summary
This connector extracts data from OneStream API and loads it directly into AWS Redshift Spectrum. 
It supports batching by custom substitution variables,

Currently supported endpoints: 
1. Data Adapters
2. SQL query
3. Run Data Management Sequence

API documentation:

https://documentation.onestream.com/1384528/Content/REST%20API/OneStream%20WebAPI%20Endpoints.html

## Importance
It solves business request to ingest data out of the OneStream database.

## Checklist

This PR:

### DIFF
--- a/src/viadot/orchestration/prefect/flows/onestream_data_adapters_to_redshift_spectrum.py
+++ b/src/viadot/orchestration/prefect/flows/onestream_data_adapters_to_redshift_spectrum.py
@@ -38,7 +38,7 @@ def onestream_data_adapters_to_redshift_spectrum(  # noqa: PLR0913
     compression: Literal["snappy", "gzip", "zstd"] | None = None,
     sep: str = ",",
     aws_config_key: str | None = None,
-    aws_credentials_secret: str | None = None,
+    credentials_secret: str | None = None,
     onestream_credentials_secret: str | None = None,
     onestream_config_key: str | None = None,
 ) -> None:
@@ -164,7 +164,7 @@ def onestream_data_adapters_to_redshift_spectrum(  # noqa: PLR0913
                 compression=compression,
                 sep=sep,
                 config_key=aws_config_key,
-                credentials_secret=aws_credentials_secret,
+                credentials_secret=credentials_secret,
             )
             if_exists = "append"  # Changed to "append" to add batches to the table.
             logger.info(
@@ -204,6 +204,6 @@ def onestream_data_adapters_to_redshift_spectrum(  # noqa: PLR0913
             compression=compression,
             sep=sep,
             config_key=aws_config_key,
-            credentials_secret=aws_credentials_secret,
+            credentials_secret=credentials_secret,
         )
         logger.info("Data processing completed successfully.")
--- a/src/viadot/orchestration/prefect/flows/onestream_sql_query_data_to_redshift_spectrum.py
+++ b/src/viadot/orchestration/prefect/flows/onestream_sql_query_data_to_redshift_spectrum.py
@@ -39,7 +39,7 @@ def onestream_sql_query_data_to_redshift_spectrum(  # noqa: PLR0913
     compression: Literal["snappy", "gzip", "zstd"] | None = None,
     sep: str = ",",
     aws_config_key: str | None = None,
-    aws_credentials_secret: str | None = None,
+    credentials_secret: str | None = None,
     onestream_credentials_secret: str | None = None,
     onestream_config_key: str | None = None,
 ) -> None:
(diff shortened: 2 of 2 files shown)


src/viadot/orchestration/prefect/flows/onestream_data_adapters_to_redshift_spectrum.py
```diff
@@ -0,0 +1,189 @@
...
+        server_url (str): OneStream server URL.
+        application (str): OneStream application name.
+        adapter_name (str): Data Adapter name to query.
+        to_path (str): Path to an S3 folder where the table will be located.
+        schema_name (str): AWS Glue catalog database name.
+        table (str): AWS Glue catalog table name.
+        workspace_name (str): OneStream workspace name. Defaults to "MainWorkspace".
+        adapter_response_key (str): Key in the JSON response that contains
+            the adapter's returned data. Defaults to "Results".
+        custom_subst_vars (dict[str, list[Any]], optional): A dictionary mapping
```
trymzet: What are "substitution variables"? It's not self-explanatory, so can you explain here the general concept before going into the details? Maybe add an example as well if it's hard to explain?

src/viadot/orchestration/prefect/flows/onestream_data_adapters_to_redshift_spectrum.py
```diff
@@ -0,0 +1,189 @@
...
+            substitution variable names to lists of possible values.
+            Values can be of any type that can be converted to strings, as they
+            are used as substitution variables in the Data Adapter. Defaults to None.
+        batch_by_subst_vars (bool): Whether to process data in batches based on
```
trymzet: This will probably make sense once you explain what subst vars are, but maybe provide an example here as well?

src/viadot/sources/onestream.py
```diff
@@ -0,0 +1,485 @@
...
+
+
+class OneStream(Source):
+    def __init__(
+        self,
+        server_url: str,
+        application: str,
+        config_key: str | None = None,
+        credentials: OneStreamCredentials | None = None,
+        api_params: dict[str, str] | None = None,
```
trymzet: Don't we already know this is an API source?

```suggestion
        params: dict[str, str] | None = None,
```

src/viadot/orchestration/prefect/tasks/onestream.py
```diff
@@ -0,0 +1,208 @@
...
+            substitution variable names to lists of possible values. Defaults to None.
+        api_params (dict[str, str], optional): API parameters. Defaults to None.
+
+    Returns:
+        pd.DataFrame: Variable combinations mapped to their data as a Pandas DataFrame.
+    """
+    if not (credentials_secret or config_key):
+        raise MissingSourceCredentialsError
+
+    credentials = get_credentials(credentials_secret)  # type: ignore
```
trymzet: Shouldn't config key be passed here too?

tests/unit/test_onestream.py
```diff
@@ -0,0 +1,464 @@
...
+        credentials=onestream_credentials,
+    )
+    assert onestream.credentials == onestream_credentials
+    assert onestream.server_url == "https://test.onestream.com"
+    assert onestream.application == "TestApp"
+    assert onestream.api_token == "test_api_token_123456"  # noqa: S105
+
+
+@patch("viadot.sources.onestream.get_source_credentials")
+def test_init_with_config_key(mock_get_creds, onestream_credentials):
```
trymzet: This isn't actually testing what it says it does, since the class doesn't use `config_key` at all and yet this test passes. This is because the return value is hardcoded to what you expect to see. So you should either rewrite this test or remove it (I don't think config key matters much currently so it's ok in this case).

BTW the way to minimize the possibility of writing such bugged tests is to use the TDD way of making sure that the test is failing first, and only then making it pass after adding the required functionality (it's even more important with agents BTW since otherwise they quite commonly make this mistake, and such bugs are quite hard to catch but can make debugging things later on a nightmare).

src/viadot/orchestration/prefect/flows/onestream_data_adapters_to_redshift_spectrum.py
```diff
@@ -0,0 +1,189 @@
+"""Extract data from OneStream Data Adapters and load it to Redshift Spectrum."""
```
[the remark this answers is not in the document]
marcinpurtak: Good point. I have added api docs also for source class and for other data adapter flow.

PR #1254, opened 2025-12-23
Add smbclient wrapper

fstajniak:
## Checklist

This PR:

### DIFF
--- a/src/viadot/sources/smb_client.py
+++ b/src/viadot/sources/smb_client.py
@@ -221,14 +221,15 @@ def _add_prefix_to_filename(filename: str, prefix: str) -> str:
 
 
 def extract_year_from_path(file_path: str) -> str | None:
-    """Extract year (4-digit, 1900-2099) from file path if present.
+    """Extract year (4-digit, 1900-2099) from directory path if present.
 
-    Searches for 4-digit numbers between 1900-2099 in the file path. If multiple
-    years are found, returns the last one (closest to the file), as it's most
-    likely the relevant year for the file.
+    Searches for 4-digit numbers between 1900-2099 in the directory part of the
+    file path only (excluding the filename). If multiple years are found, returns
+    the last one (closest to the file), as it's most likely the relevant year
+    for the file.
 
     Args:
-        file_path (str): The file path to search for years.
+        file_path (str): The file path to search for years in directories only.
 
     Returns:
         str | None: The extracted year as a string, or None if no valid year found.
@@ -240,10 +241,16 @@ def extract_year_from_path(file_path: str) -> str | None:
         '2024'
         >>> extract_year_from_path("data/file.xlsx")
         None
+        >>> extract_year_from_path("2024/data/file1926.xlsx")
+        '2024'
     """
-    # Match 4-digit year (1900-2099) in the path
+    directory_path = str(Path(file_path).parent)
+    if directory_path == ".":
+        return None
+
+    # Match 4-digit year (1900-2099) in the directory path only
     year_pattern = r"\b(19\d{2}|20[0-9]{2})\b"
-    matches = re.findall(year_pattern, file_path)
+    matches = re.findall(year_pattern, directory_path)
     if matches:
         # Return the last year found (closest to the file in the path)
         return matches[-1]


src/viadot/sources/smbclient_wrapper.py
```diff
@@ -0,0 +1,1824 @@
...
+            aws_credentials=aws_credentials,
+            timeout=timeout,
+            prefix_levels_to_add=prefix_levels_to_add,
+        )
+    except SMBInvalidFilenameError:
+        logger.warning(f"Skipping file with invalid filename: {remote_path}")
+        return None
+
+
+def stream_smb_files_to_s3(  # noqa: PLR0913
```
Rafalz13: Maybe it can be part of the connector class, or it should be moved to tasks. To be able to easily run it. Currently, this function is in the file, but it's not accessible importing class.
Consider moving it into task or even flow.

The same with the rest functions created outside class

src/viadot/sources/smbclient_wrapper.py
```diff
@@ -0,0 +1,1824 @@
...
+            # Example: "file.xlsx       A    19489  Thu Apr 10 14:41:14 2025"
+            match = re.match(
+                r"^(.+?)\s+([A-Z])\s+(\d+)\s+(\w+)\s+(\w+)\s+(\d+)\s+(\d+:\d+:\d+)\s+(\d+)$",
+                stripped,
+            )
+            if match:
+                name = match.group(1).strip()
+                type_char = match.group(2)
+                size = int(match.group(3))  # File size in bytes
+                # day_name = match.group(4)  # Not used
```
Rafalz13: remove if this is not used

src/viadot/sources/smbclient_wrapper.py
fstajniak: I dont think it is necessary because new connector (class) provides 3 functions:

- list_directory (wraps the get_hybrid_listing_with_fallback function) -> listing files on an SMB server
- download_file (wraps the download_file_from_smb_with_retry function) -> downloading files locally from SMB server
- stream_files_to_s3 (wraps the stream_smb_files_to_s3 function) -> streaming using STDOUT->STDIN from SMB server to S3

src/viadot/sources/smb_client.py
```diff
@@ -0,0 +1,1734 @@
...
+    if len(paths) == 1:
+        logger.info(f"Successfully downloaded file to: {paths[0]}")
+    else:
+        logger.info(
+            f"Successfully extracted {len(paths)} files from ZIP: {remote_path}"
+        )
+    return paths
+
+
+def _stream_file_from_smb_to_s3(  # noqa: C901, PLR0912, PLR0915
```
trymzet: Looks like you're hardcoding the destination in the source, which is not the way to go about it, since it's a lot of code that has nothing to do with the source connector (and also seems like a lot of duplication between what is here and in the `S3` source, whereas all S3 logic should be in the S3 connector).

Instead, the connector should work with generators, so that the task/destination can request the data one chunk at a time from the connector, eg.:

```python
# The source generates data in chunks, which can be consumed as required.
def to_df():
    for chunk in df_chunks:
        yield df_chunk
```

then eg. in the flow:

```python
smb = SMB()
s3 = S3()

@task
def smb_to_s3(to_df_params, from_df_params):
    # Process one chunk of data (file?) at a time.
    for chunk in smb.to_df(**to_df_params):
        s3.from_df(chunk, **from_df_params)
```
This way you generate and upload one chunk at a time without having to hardcode the destination in the source connector.

src/viadot/sources/smb_client.py
```diff
@@ -0,0 +1,1734 @@
...
+class SMBClient(Source):
+    """Lightweight `Source` connector built around helpers in this module.
+
+    This class does **not** introduce new complex logic (such as full tree scanning),
+    it only wraps the existing helper functions:
+
+    - `get_listing_with_shallow_first` - shallow-first directory listing with recursive
+    expansion,
+    - `download_file_from_smb` - file download with optional ZIP extraction,
+    - `stream_smb_files_to_s3` - streaming files directly to S3.
```
trymzet: The connector still contains all this complexity whether you define the logic here or 10 lines above. I guess you could rewrite this as listing the functionalities of the connector:

```suggestion
   Functionality:
   - recursively listing directories
   - file download with optional ZIP extraction
```

src/viadot/sources/smb_client.py
```diff
@@ -0,0 +1,1734 @@
...
+                dynamic_date_symbols=dynamic_date_symbols,
+                dynamic_date_format=dynamic_date_format,
+                dynamic_date_timezone=dynamic_date_timezone,
+            )
+
+        if should_include:
+            paths.append(self.path)
+
+        # Recursively check children
+        for child in self.children:
```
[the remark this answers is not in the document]
fstajniak: Directories can contain files and that's why we need to check each files inside that directory (recursively)

Rafalz13 (approved):

PR #1255, opened 2025-12-30
ECB Currencies Connector

DominikTyrala:


src/viadot/sources/ecb.py
```diff
@@ -0,0 +1,183 @@
...
+from viadot.exceptions import APIError
+from viadot.sources.base import Source
+from viadot.utils import (
+    add_viadot_metadata_columns,
+    handle_api_response,
+    validate,
+)
+
+
+class ECB(Source):
```
trymzet: It seems this connector does not handle the ECB API but instead only one specific endpoint, the exchange rates one, correct? If so, this should be called `ECBExchangeRates` and the `url` should actually be hardcoded to `https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml`.

Alternatively, we could leave this as ECB, but then you need to refactor the connector to work for other endpoints as well (so parsing function would either need to be rewritten to be generic or you'd have one parsing function per endpoint - with only one function currently implemented, called eg. `_parse_exchange_rates_xml()`).

trymzet (changes requested):
Added my comments

PR #1276, opened 2026-03-12
🚀 Add new param `gh_actor` to transform and catalog flow

AndriiKachan:
this also introduces new viadot version 2.4.2


marcinpurtak (approved):

trymzet:
@AndriiKachan what is this param for? It doesn't seem like tracking sth like this belongs to the flow, but rather should be done in the relevant GitHub workflow.

AndriiKachan:
> @AndriiKachan what is this param for? It doesn't seem like tracking sth like this belongs to the flow, but rather should be done in the relevant GitHub workflow.

Hello @trymzet , it is used by our client internal CI workflow which runs `dbt_rerun` deployment, which uses this flow. Basically, we created an Action and can see who triggered/requested its run on Prefect. Please align with @burzecj , as its his feature.

burzecj:
We were required to create a GitHub Actions workflow so that the werfen team could rerun dbt models without access to the prod Prefect account. Since we were using a prod user within Prefect to trigger deployments, we had to add annotations to make it clear in the Prefect logs who initiated a given deployment. This parameter is only passed when invoking `dbt-rerun` via the GitHub Actions workflow.

trymzet:
Makes sense, though it looks like this could be better solved by eg. running the deployment with a custom tag (which would be passed in the GitHub workflow), rather than modifying the flow with a param that's completely unrelated to its logic.

But at least now the reason for this feature is discoverable in the repo so we can refactor this in the future if required.

burzecj:
> Makes sense, though it looks like this could be better solved by eg. running the deployment with a custom tag (which would be passed in the GitHub workflow), rather than modifying the flow with a param that's completely unrelated to its logic.
> 
> But at least now the reason for this feature is discoverable in the repo so we can refactor this in the future if required.

If there's an option to update the tag when the deployment is triggered, we can definitely do that. 
When we were setting up this GitHub workflow, this solution seemed like the simplest option at first, but we can certainly improve it.

@AndriiKachan could you create a new GH Issue regarding improving this approach and put it to DE/DevOps backlog? 
I'd appreciate it

trymzet:
Yes, you can set tags at the deployment run level (`prefect deployment run my_deployment --tag my_tag`). Sounds great.

AndriiKachan:
Issue created:
- https://github.com/dyvenia/werfen-project/issues/4254

PR #1303, opened 2026-04-07
luma_url default value removed

k-wolski:
## Summary
In order to enable WAF for the projects, we should avoid setting default values to localhost. Prefect CLI during deployment registration is not allowing such flows, even though we are setting the `luma_url` to meaningful value.

Example of blocked request:

image


Rafalz13 (dismissed):

k-wolski:
> @k-wolski Seems like a bug in Prefect? It should definitely not be breaking because you set some param default to certain string value

It's not Prefect blocking that, but it's WAF - web application firewall. The behavior is different between runtime and flow registration. During registration, Prefect CLI sends a POST request to the Prefect server API (`/api/deployments/`) with a full JSON description of the flow: its name, parameters, schedule, and the entire function's docstring as the description field.

WAF inspects the registration request body from that call, and the docstring still contains `Defaults to "http://localhost:8000".`, and that's the problem.

trymzet (approved):

PR #1305, opened 2026-04-14
Pyarrow round 2

DominikTyrala:
## Checklist

This PR:

- [ ] contains a summary of the changes (the "what")

### DIFF
--- a/src/viadot/orchestration/prefect/tasks/sap_rfc.py
+++ b/src/viadot/orchestration/prefect/tasks/sap_rfc.py
@@ -120,7 +120,7 @@ def sap_rfc_to_df(  # noqa: PLR0913
     sap.query(query)
     logger.info("Downloading data from SAP to a DataFrame...")
     logger.debug(f"Running query: \n{sap.sql}.")
-    logger.debug(f"Dominik Run Metadata Updated")
+    logger.debug(f"Dominik Run Metadata Updated. Updated Creds")
 
     arrow_table = sap.to_arrow(tests=tests)
 
--- a/src/viadot/orchestration/prefect/tasks/sap_rfc.py
+++ b/src/viadot/orchestration/prefect/tasks/sap_rfc.py
@@ -120,7 +120,6 @@ def sap_rfc_to_df(  # noqa: PLR0913
     sap.query(query)
     logger.info("Downloading data from SAP to a DataFrame...")
     logger.debug(f"Running query: \n{sap.sql}.")
-    logger.debug(f"Dominik Run Metadata Updated. Updated Creds")
 
     arrow_table = sap.to_arrow(tests=tests)
 
--- a/src/viadot/orchestration/prefect/tasks/sap_rfc.py
+++ b/src/viadot/orchestration/prefect/tasks/sap_rfc.py
@@ -121,10 +121,13 @@ def sap_rfc_to_df(  # noqa: PLR0913
     logger.info("Downloading data from SAP to a DataFrame...")
     logger.debug(f"Running query: \n{sap.sql}.")
 
-    df = sap.to_df(tests=tests)
+    arrow_table = sap.to_arrow(tests=tests)
 
-    if not df.empty:
+    if arrow_table.num_rows > 0:
         logger.info("Data has been downloaded successfully.")
-    elif df.empty:
+    else:
         logger.warn("Task finished but NO data was downloaded.")
-    return df
+
+    return arrow_table.to_pandas(
+        types_mapper=None, split_blocks=True, self_destruct=True
+    )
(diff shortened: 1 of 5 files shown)


src/viadot/sources/sap_rfc.py
```diff
@@ -789,3 +797,306 @@ def to_df(self, tests: dict | None = None) -> pd.DataFrame:  # noqa: C901, PLR09
             validate(df=df, tests=tests)
 
         return df
+
+    def split_wa_records(
```
winiar93, line 801: here are creating new list from reposne response["DATA"] which are in memory already. so we have two almost same data at once

This one uses list comprehension, 
could we consider to use generator like in `_gen_split` like in to_df ?
we could reduce peak memory usage

src/viadot/sources/sap_rfc.py
```diff
@@ -789,3 +797,306 @@ def to_df(self, tests: dict | None = None) -> pd.DataFrame:  # noqa: C901, PLR09
...
+                if alias_map:
+                    for row in result_rows:
+                        # create a list to avoid modifying dict during iteration
+                        for orig_col, alias_col in list(alias_map.items()):
+                            if orig_col in row and alias_col != orig_col:
+                                row[alias_col] = row.pop(orig_col)
+
+                # Drop columns not selected by the user
+                selected_set = set(columns_aliased)
+                for i in range(len(result_rows)):
```
winiar93, line 1062: in 1054 and in 1062 there is iteration over same rows

there are three loops
1 - alias map
2 - removing cols
3 - remove cols from client_side_filters

maybe is it possible to iterate once instead of three times over same data ?

src/viadot/sources/sap_rfc.py
```diff
@@ -789,3 +797,306 @@ def to_df(self, tests: dict | None = None) -> pd.DataFrame:  # noqa: C901, PLR09
...
+                        # Align row count with the first column_batch
+                        if row_index == 0:
+                            row_index = len(records)
+                            if row_index == 0:
+                                logger.warning(
+                                    f"Empty output was generated for column batch {column_batch} in columns {fields}."
+                                )
+                                # nothing to add for this column_batch
+                            else:
+                                rows_list = [{} for _ in range(row_index)]
```
winiar93, line 1018: could we avoid this preallocation? For a dataset of 2 000 000 rows, sys.getsizeof(rows_list) shows ~17 mb allocated just for the empty dict before any actual data is inserted

Image

winiar93:
Just to confirm that I understand it correctly, this connector loads the entire sap reposioine into memory at once then processes it ?

if so when we are going to load a lot of data we going to have:
- response[DATA] which is full response from sap
- split_wa_records which is scopy of same data after splitting
- rows_list another data representation

What if we fetched data in row chunks (using LIMIT with OFFSET ) and inserted each chunk directly into the database? In that way we dont need to hold the entire dataset in memory at once.

trymzet:
> Just to confirm that I understand it correctly, this connector loads the entire sap reposioine into memory at once then processes it ?
> 
> if so when we are going to load a lot of data we going to have:
> 
> * response[DATA] which is full response from sap
> * split_wa_records which is scopy of same data after splitting
> * rows_list another data representation
> 
> What if we fetched data in row chunks (using LIMIT with OFFSET ) and inserted each chunk directly into the database? In that way we dont need to hold the entire dataset in memory at once.

pyrfc doesn't support sessions, so we cannot simply use LIMIT and OFFSET, as data could change between the calls (inserts, deletes, schema changes, etc.). Hopefully the custom connector can add this feature one day? Then we could  `yield chunk` everywhere so that extraction, processing, and ingestion are all done in chunks.

src/viadot/orchestration/prefect/tasks/sap_rfc.py
```diff
@@ -115,11 +115,15 @@ def sap_rfc_to_df(  # noqa: PLR0913
     sap.query(query)
     logger.info("Downloading data from SAP to a DataFrame...")
     logger.debug(f"Running query: \n{sap.sql}.")
+    logger.debug(f"Dominik Run Metadata Updated")
```
trymzet:
```suggestion
```

src/viadot/sources/sap_rfc.py
DominikTyrala: ok, let me try to implement different approach.

src/viadot/sources/sap_rfc.py
DominikTyrala: for sure there is better approach, but I would need more time to create something more sophisticated. so I would propose to go with current implementation.

DominikTyrala:
@winiar93 @trymzet - I agree with your comments as replied above. However it seems that these are bit more complicated and requires another round of testing. As this seems to work and is already tested I suggest to merge this, and in the next commit the proposed improvements. Agreed on this with @k-wolski

src/viadot/orchestration/prefect/tasks/sap_rfc.py
DominikTyrala: done

trymzet (approved):

PR #1307, opened 2026-04-16
✨ Add Perspective dbt ingestion to T&C flow

trymzet:
## Summary

Adds ability to ingest dbt run results artifact into Perspective Catalog in the Transform and Catalog flow.

This also improves the CI workflow to check that `uv.lock` is always up to date.

## Checklist

This PR:


src/viadot/orchestration/prefect/tasks/perspective.py
```diff
@@ -0,0 +1,68 @@
...
+    dry_run: bool = False,
+) -> subprocess.CompletedProcess:
+    """Upload test results to Perspective.
+
+    Args:
+        perspective_api_url (str | None): URL of the Perspective catalog to which test
+            results will be sent. Defaults to None.
+        perspective_api_token (str | None): Token used for communication with
+            Perspective. Defaults to None.
+        target_path (str | Path | None): Path to dbt target directory in which
```
marcinpurtak, line 26: If target_path is a source of `run_results.json` generated during dbt run, maybe change it to `from_path` or `metadata_path` to keep the parameters name consistent with `luma_ingest_task`.

trymzet: I think it wasn't done correctly in Luma actually, since dbt uses the term "target" and has the `--target-path` param itself, so I think it's better to just reuse dbt terminology instead of inventing something

marcinpurtak (approved):

PR #1312, opened 2026-05-12
✨ Add support for partial parse in transform_and_catalog flow

AndriiKachan:
- Added a shared dbt `ArtifactStore` abstraction for `manifest.json`, `partial_parse.msgpack`, and historical `run_results.json` artifacts.
- Replaced separate manifest/run-results storage params in `transform_and_catalog` with 1 root `artifact_store_path`.
- Added partial-parse download before dbt runs
- Updated to read manifests through `ArtifactStore`.
- Added tests for artifact path building, partial-parse download, and run-results upload.


AndriiKachan:
Discussed changes added @trymzet , please check

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
```diff
@@ -348,23 +389,39 @@ def transform_and_catalog(  # noqa: PLR0913, PLR0915, C901 | Complexity complain
...
+        if state_store_credentials_secret
+        else None
+    )
+    manifest_store_credentials = (
+        get_credentials(
+            manifest_store_credentials_secret or run_results_storage_credentials_secret
+        )
+        if manifest_store_credentials_secret or run_results_storage_credentials_secret
+        else None
+    )
```
trymzet: We don't want these fallbacks; the'yre different params for different purpose
```suggestion
    state_store_credentials = get_credentials(state_store_credentials_secret)
    manifest_store_credentials = get_credentials(manifest_store_credentials_secret)
```

src/viadot/orchestration/prefect/flows/transform_and_catalog.py
```diff
@@ -458,7 +515,7 @@ def transform_and_catalog(  # noqa: PLR0913, PLR0915, C901 | Complexity complain
             from_path=run_results_file_path,
             to_path=run_results_storage_path,
             config_key=run_results_storage_config_key,
```
trymzet: I'd get rid of that as it's not really used
```suggestion
```

trymzet (changes requested):
I've given this some thought, and I think this and run results as well should be integrated into the `ManifestStore` class and the class itself renamed into `ArtifactStore`, with methods for fetching the other two files.

This is because I don't like how manifest, run_results, and partial_parse all use different code, params, and logic, even though they're pretty much the same thing (we're downloading/uploading a file to S3). This adds a bunch of complexity to this flow, which is already too large and very messy, so we need to find a way to simplify this.

We should only be accepting `artifact_store_type`, `artifact_store_credentials_secret`, and `artifact_store_path` in the flow params, and just use those to figure out all the rest (eg. manifest_path becomes `Path(artifact_store_path) / "manifest.json")`. Also this way you don't need to hardcode S3 like you do in your task currently - the artifact store type is passed as param.

Let's maybe meet to discuss this?

AndriiKachan:
Hello @trymzet , please check.
For me, it's now too many changes honestly.. Thoughts?

Also, should `manifest_store.py‎` stay with backward compatibility or better remove it?

trymzet (approved):

PR #1314, opened 2026-05-14
🔥 Remove old Rye files

trymzet:
## Summary

Removes old Rye files, as we moved to `uv` a while ago.

Not sure how the container is used; this approach requires user to activate the uv venv or run commands that utilize dev dependencies, such as `mkdocs`, with `uv run` (`uv run mkdocs build` etc.) which I guess should be fine?


k-wolski (approved):
These changes are fine. What we can consider is to find a way for local development of SAP connector without docker container. Right now it's the only purpose of these images, as viadot in other projects is used only as python dependency.

trymzet:
@k-wolski the setup for local development is already documented [in the docs](https://viadot.docs.dyvenia.com/getting_started/#os-dependencies), so this can be easily set up if someone needs it. The docker containers are already optional and only a convenience.

PR #1323, opened 2026-05-29
🐛 Fixes for SLA Monitor flow

marcinpurtak:
## Summary

This PR fixes issues with the SLA Monitor.
What was fixed:
- Wrong node-status.json fields were used
- SMTP credentials handling
- Updates of node-status.json
- Reset of _sla_breach_notification_sent even if sla was breached
- Changed order of checks to skip early if node is not a model
- Modified node status write function so it doesn't require status to be passed

## Checklist

This PR:

- [ ] contains a summary of the changes (the "what")

### DIFF
--- a/src/viadot/orchestration/prefect/flows/dbt_sla_monitor.py
+++ b/src/viadot/orchestration/prefect/flows/dbt_sla_monitor.py
@@ -162,6 +162,9 @@ def sla_monitor(
             continue
 
         if node.get("TYPE") != "model":
+            prefect_logger.debug(
+                f"Node '{node_name}' is not a model; skipping SLA check."
+            )
             logger.debug(f"Node '{node_name}' is not a model; skipping SLA check.")
             continue
 
@@ -172,6 +175,9 @@ def sla_monitor(
             continue
 
         if node.get("fresh_until") is None:
+            prefect_logger.warning(
+                f"No 'fresh_until' timestamp for node '{node_name}'; cannot validate SLA."
+            )
             logger.warning(f"No SLA found for node '{node_name}'; cannot validate.")
             continue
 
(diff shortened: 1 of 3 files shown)


src/viadot/orchestration/prefect/flows/dbt_sla_monitor.py
```diff
@@ -155,27 +168,36 @@ def sla_monitor(
             # Reset SLA breach notification flag on successful runs to allow future
             # breach notifications.
             store.write(
-                node_state={node_name: {"_sla_breach_notification_sent": False}}
+                node_state={
+                    "table_name": node_name,
+                    "status": "success",
```
trymzet: We shouldn't be updating the status in this flow
```suggestion
```

src/viadot/orchestration/prefect/flows/dbt_sla_monitor.py
```diff
@@ -94,7 +99,13 @@ def _handle_breached_node(
         recipients=owners,
         smtp_credentials_secret=smtp_credentials_secret,
     )
-    store.write(node_state={node_name: {"_sla_breach_notification_sent": True}})
+    store.write(
+        node_state={
+            "table_name": node_name,
+            "status": node["status"],
```
trymzet: We shouldn't be updating the status in this flow
```suggestion
```

trymzet (changes requested):

src/viadot/orchestration/prefect/flows/dbt_sla_monitor.py
marcinpurtak: I had to put it. Otherwise the flow was crushing. Seems like we expect to allways put the status in the update function.

marcinpurtak: I'm going to change the update function so it doesn't request status by default.

src/viadot/orchestration/prefect/flows/dbt_sla_monitor.py
marcinpurtak: I'm going to change the update function so it doesn't request status by default.

trymzet (dismissed):

trymzet (approved):

Rafalz13 (approved):

PR #1344, opened 2026-07-13
Extend `SMB` connector to handle file paths

fstajniak:
Adds fetch_and_store() to the SMB connector for reading files from known paths without directory scanning. Fixes logging when streaming ZIP files with zip_inner_file_regexes.


Rafalz13 (commented):
Looks good, please update tests to pass

Rafalz13 (approved):

PR #1359, opened 2026-07-15
added logs to sap rfc task

winiar93:
## Summary
Closes: [4553](https://github.com/dyvenia/werfen-project/issues/4553)

Add detailed logging to the `sap_rfc_to_df` task to improve observability and ease debugging of failed Prefect flow runs.

## Importance

The task previously provided minimal insight into what was happening during execution, making it hard to diagnose failures or performance issues.
This PR adds:
- Logging of the query after dynamic date parsing
- Measure time around sap query / download step
- Row / column counts
- A `try/except` block around the query execution and data fetch
- DataFrame shape and memory usage after conversion from Arrow to pandas

image

image

These changes make it significantly easier to pinpoint where and why a flow run failed

## Checklist

This PR:

- [x] links the relevant issue with the `Closes #<issue_number>` phrase (the "why")


trymzet:
@winiar93 seems like most/all of these should be added to the connector? The Prefect tasks/flows are supposed to be minimal wrappers around the connectors (so actually this task is already far too complicated for what it's doing, as it should just be calling `sap.to_df()`).

winiar93:
> @winiar93 seems like most/all of these should be added to the connector? The Prefect tasks/flows are supposed to be minimal wrappers around the connectors (so actually this task is already far too complicated for what it's doing, as it should just be calling `sap.to_df()`).

ok, code adjusted. Logs moved to source class

src/viadot/sources/sap_rfc.py
```diff
@@ -790,12 +809,24 @@ def to_df(self, tests: dict | None = None) -> pd.DataFrame:  # noqa: C901, PLR09
                 for col in client_side_filter_cols_aliased
                 if col not in self.select_columns_aliased
             ]
+            if cols_to_drop:
+                self.logger.debug(
+                    f"Dropping helper column(s) used only for client-side filtering: {cols_to_drop}."
```
Rafalz13: I'm not sure if `client-side` is correct. It can be anything - not only client-side

Rafalz13 (commented):
You mention a list of columns in loggers a few times. Is that needed?

src/viadot/sources/sap_rfc.py
winiar93: Adjusted, done

winiar93:
> You mention a list of columns in loggers a few times. Is that needed?

Yes, I think its usefull, they are short, so they won't fload the logs.
For me, now, it's much easier to understand what is going on.

Rafalz13 (approved):

PR #1362, opened 2026-07-20
🔧  adjusted `name` param in `matomo_to_redshift_spectrum` flow

damianbugajDE:
image


trymzet (approved):

damianbugajDE:
Checked DEV run with `custom` image and it pulled same amount of data as the one from `master` image pulled on PROD
DEV:
image

---

PROD:
image

PR #1372, opened 2026-08-05
🔧 Update AWS creds handling

akachandyvenia:
## Summary

Fix AWS credential loading for IRSA-backed Prefect blocks that do not contain static access keys.

## Changes

- Handle a missing `aws_secret_access_key` without raising `AttributeError`.
- Preserve support for static AWS credentials.


trymzet:
I thought you don't need to pass any credentials if you're using IRSA? Or is there some scenario where access key is needed, but secret key is not? Or is the region still required? But then it might be better to have a `AWS_REGION` env var configured at the infra level.

akachandyvenia:
> I thought you don't need to pass any credentials if you're using IRSA? Or is there some scenario where access key is needed, but secret key is not? Or is the region still required? But then it might be better to have a `AWS_REGION` env var configured at the infra level.

I agree, please check @trymzet

src/viadot/orchestration/prefect/utils.py
```diff
@@ -959,11 +959,26 @@ def _get_aws_credentials(
...
+        aws_secret_access_key_block = aws_credentials_block.aws_secret_access_key
+        aws_secret_access_key = (
+            aws_secret_access_key_block.get_secret_value()
+            if aws_secret_access_key_block
+            else None
+        )
+
+        if bool(aws_access_key_id) != bool(aws_secret_access_key):
+            msg = "AWS access key ID and secret access key must be provided together."
+            raise ValueError(msg)
```
trymzet: I think this is more reabable
```suggestion
        if not (aws_access_key_id and aws_secret_access_key):
            msg = "AWS access key ID and secret access key must be provided together."
            raise ValueError(msg)
```

akachandyvenia: Outdated

src/viadot/sources/redshift_spectrum.py
```diff
@@ -50,19 +50,36 @@ class RedshiftSpectrumCredentials(BaseModel):
...
+        if not (
+            profile_credential
+            or direct_credential
+            or machine_credential
+            or partial_machine_credential
+        ):
+            msg = "AWS credentials must be omitted, or specify `region_name` only, "
+            msg += "`profile_name` and `region_name`, or `aws_access_key_id`, "
+            msg += "`aws_secret_access_key`, and `region_name`."
             raise CredentialError(msg)
```
trymzet, line 83: this is never reached so I'd remove this block

src/viadot/sources/s3.py
```diff
@@ -57,11 +58,24 @@ def is_configured(cls, credentials: dict[str, Any]) -> dict[str, Any]:
...
+        if not (
+            profile_credential
+            or direct_credential
+            or machine_credential
+            or partial_machine_credential
+        ):
+            msg = "AWS credentials must be omitted, or specify `region_name` only, "
+            msg += "`profile_name` and `region_name`, or `aws_access_key_id`, "
+            msg += "`aws_secret_access_key`, and `region_name`."
             raise CredentialError(msg)
```
trymzet, line 79: this is never reached so I'd remove this block

src/viadot/sources/redshift_spectrum.py
akachandyvenia: This block is reachable for incomplete explicit credentials, such as `region_name` with only `aws_access_key_id` - it prevents malformed configurations from falling through to boto3, so it should remain

src/viadot/sources/s3.py
akachandyvenia: This block is reachable for incomplete explicit credentials, such as `region_name` with only `aws_access_key_id` - it prevents malformed configurations from falling through to boto3, so it should remain

trymzet (approved):

PR #1374, opened 2026-08-11
🤖 Add `salesforces_to_redshift_spectrum` flow

winiar93:
## Summary
Added flow for salesforces to redshift spectrum and applied small required adjustments for salesforce connector.

Salesforce is discontinuing support for the OAuth 2.0 Username-Password Flow 
for connected apps. The authentication method is being changed to use 
Consumer Key + Consumer Secret (OAuth 2.0 Client Credentials Flow) instead.

## Importance

Client need to load data from salesfore

## Checklist

This PR:

- [x] links the relevant issue with the `Closes #<issue_number>` phrase (the "why")


src/viadot/orchestration/prefect/tasks/salesforce.py
```diff
@@ -11,10 +11,9 @@
 @task(retries=3, log_prints=True, retry_delay_seconds=10, timeout_seconds=60 * 60)
 def salesforce_to_df(
     config_key: str | None = None,
-    azure_key_vault_secret: str | None = None,
+    salesforce_credentials_secret: str | None = None,
     env: str | None = None,
     domain: str | None = None,
-    client_id: str | None = None,
```
trymzet, line 17: Why is this removed? Isn't it still required when using simple auth?

src/viadot/sources/salesforce.py
```diff
@@ -14,21 +15,17 @@
...
+            authenticate with Salesforce.
+        consumer_secret (str): The client secret of the connected app used to
+            authenticate with Salesforce.
     """
 
-    username: str
-    password: str
-    token: str
+    consumer_key: str
+    consumer_secret: str
```
trymzet, line 33: Is simple auth no longer supported by Simple Salesforce? If it still is, token auth support should be added as a new feature instead of replacing existing behavior.

src/viadot/sources/salesforce.py
```diff
@@ -61,42 +58,84 @@ def __init__(
...
+        self.instance_url = instance_url
+
+        if self.env in ("DEV", "QA") and not self.instance_url:
+            message = (
+                f"'instance_url' is required when env='{self.env}'. "
+                "Provide it explicitly, e.g. instance_url='https://your-domain.my.salesforce.com'."
+            )
+            raise ValueError(message)
+
+        self.token = self.generate_token()
```
trymzet, line 91: This should not be called at init time unless token auth is the only supported auth method

src/viadot/sources/salesforce.py
```diff
@@ -61,42 +58,84 @@ def __init__(
...
+        if not client_secret:
+            message = "'consumer_secret' is required to generate a token."
+            raise CredentialError(message)
+
+        if self.env == "PROD":
+            base_url = "https://login.salesforce.com"
+        else:
+            if not self.instance_url:
+                message = f"'instance_url' is required to generate a token for env='{self.env}'."
+                raise ValueError(message)
```
trymzet: This check should be done at init time. Also, set instance url at init too `self.instance_url = instance_url if env != "PROD" else "https://login.salesforce.com"`. Then you don't need this block here and can just do
```python
token_url = f"{self.instance_url}/services/oauth2/token"
```

src/viadot/sources/salesforce.py
```diff
@@ -123,11 +162,11 @@ def to_df(
             columns_str = ", ".join(columns) if columns else "FIELDS(STANDARD)"
             query = f"SELECT {columns_str} FROM {table}"  # noqa: S608
 
-        data = self.salesforce.query(query).get("records")
+        data = self.salesforce.query_all(query).get("records") or []
```
trymzet: I think we should be moving to using generators where possible to better control memory use. You can instead do:
```python
from itertools import batched

data = self.salesforce.query_all_iter(query)

# Checking emptiness is a bit trickier with iterators; I kept it simple here for demonstration
if not data:
    self._handle_if_empty(if_empty)

# Yield data in chunks to control memory use.
for chunk in batched(data, chunk_size):  # add chunk_size as param to to_df() and in the upstream task and flow. Probably default to 50000?
    yield pd.DataFrame(chunk).drop("attributes")  # do post-processing per chunk, not per row
```

tests/unit/test_salesforce.py
```diff
@@ -69,10 +75,21 @@ def mock_sf_instance(mocker):
     return mock_sf_instance
 
 
+@pytest.fixture
+def salesforce_instance(mock_sf_instance, mock_generate_token):  # noqa: ARG001
+    """Fixture providing a Salesforce instance with mocked auth and API client."""
+    return Salesforce(
+        credentials=variables["credentials"],
+        instance_url="https://test.salesforce.com",
+    )
```
trymzet: There's already `mock_sf_instance`, so either use that one as-is or modify it as needed.

trymzet (changes requested):

src/viadot/sources/salesforce.py
winiar93: it is salesforce recommendation 
https://help.salesforce.com/s/articleView?id=000886201&type=1


Also, I never had the chance to use user/passwd to authh, it was always via connected apps with consumer key + secret

src/viadot/sources/salesforce.py
winiar93: added

src/viadot/sources/salesforce.py
winiar93: added

tests/unit/test_salesforce.py
winiar93: adjusted, done

src/viadot/orchestration/prefect/tasks/salesforce.py
winiar93: auth params changed

src/viadot/orchestration/prefect/flows/salesforce_to_redshift_spectrum.py
```diff
@@ -69,34 +73,45 @@ def salesforce_to_redshift_spectrum(  # noqa: PLR0913
...
         salesforce_table (str, optional): The name of the Salesforce
             table to extract data from. Defaults to None.
         columns (list[str], optional): The list of columns to extract from
             the Salesforce table. Defaults to None.
+        chunk_size (int, optional): The number of rows to be fetched in each chunk.
+            Defaults to None.
     """
-    df = salesforce_to_df(
+    logger = get_run_logger()
+    logger.setLevel(logging.DEBUG)
```
trymzet: You shouldn't do that here. Just use `logger.debug()` when you need to print debug-level logs, and then use relevant Prefect configuration to show DEBUG-level logs in the UI. You can do that by overriding Prefect worker env vars by specifying relevant `job_variables` in your deployment run:
```json
{
  "RUNTIME__LOG_LEVEL": "DEBUG",
  "PREFECT_LOGGING_LEVEL": "DEBUG",
  "PREFECT_LOGGING_ROOT_LEVEL": "DEBUG",
  "PREFECT_LOGGING_LOGGERS_PREFECT_EXTRA_LEVEL": "DEBUG"
}
```
(probably only 1 or 2 of these are required but I never figured out which)

src/viadot/orchestration/prefect/tasks/salesforce.py
```diff
@@ -51,6 +58,10 @@ def salesforce_to_df(
         config_key=config_key,
         env=env,
         domain=domain,
+        instance_url=instance_url,
+    )
+    return list(
+        salesforce.to_df(
+            query=query, table=table, columns=columns, chunk_size=chunk_size
+        )
```
trymzet: The goal is to process data in chunks. `list()` defeats that purpose. This task should also yield chunks

src/viadot/sources/salesforce.py
```diff
@@ -136,49 +135,49 @@ def generate_token(self) -> dict:
...
             query = f"SELECT {columns_str} FROM {table}"  # noqa: S608
 
-        data = self.salesforce.query_all(query).get("records") or []
+        records_iter = self.salesforce.query_all_iter(query)
 
-        # Remove metadata from the data
-        for record in data:
-            record.pop("attributes", None)
+        chunk_size = chunk_size or DEFAULT_CHUNK_SIZE
+        got_any_data = False
```
trymzet: maybe `is_empty`?

src/viadot/sources/salesforce.py
```diff
@@ -136,49 +135,49 @@ def generate_token(self) -> dict:
         self.logger.info("Successfully generated Salesforce access token.")
         return token_data
 
-    @add_viadot_metadata_columns
```
trymzet, line 100: These columns still need to be added. Question is, can you improve the decorator to also work with `to_df()` which are generators or do we need some workaround solution for this PR and we'll add the proper solution later.

trymzet (changes requested):

src/viadot/sources/salesforce.py
winiar93: Decorator was updated

src/viadot/orchestration/prefect/flows/salesforce_to_redshift_spectrum.py
winiar93: ok, removed

src/viadot/orchestration/prefect/tasks/salesforce.py
winiar93: done

src/viadot/sources/salesforce.py
winiar93: renamed, done

trymzet (changes requested):
Viadot metadata columns still left

winiar93:
> Member

It was added

src/viadot/sources/salesforce.py
```diff
@@ -61,85 +63,123 @@ def __init__(
...
-        df = pd.DataFrame(data)
+        chunk_size = chunk_size or DEFAULT_CHUNK_SIZE
+        is_empty = False
+        # Yield data in chunks to control memory use.
+        for chunk in batched(records_iter, chunk_size):
+            is_empty = True
+            yield pd.DataFrame(chunk).drop(columns=["attributes"], errors="ignore")
 
-        if df.empty:
+        if not is_empty:
```
trymzet: Make it logical:
```suggestion
        if is_empty:
```

src/viadot/sources/salesforce.py
```diff
@@ -61,85 +63,123 @@ def __init__(
...
-        data = self.salesforce.query(query).get("records")
-
-        # Remove metadata from the data
-        for record in data:
-            record.pop("attributes")
+        records_iter = self.salesforce.query_all_iter(query)
 
-        df = pd.DataFrame(data)
+        chunk_size = chunk_size or DEFAULT_CHUNK_SIZE
+        is_empty = False
```
trymzet:
```suggestion
        is_empty = True
```

src/viadot/sources/salesforce.py
```diff
@@ -61,85 +63,123 @@ def __init__(
...
+        # Yield data in chunks to control memory use.
+        for chunk in batched(records_iter, chunk_size):
+            is_empty = True
```
trymzet:
```suggestion
            is_empty = False
```

trymzet (changes requested):

trymzet (approved):

